PR #3, opened 2019-07-25
Refs #8646. Added endpoints for Algorithm, Job and Result API. ISSUE #2

haimasree:
Closes: #2

### DIFF
--- a/gcapi/gcapi.py
+++ b/gcapi/gcapi.py
@@ -38,11 +38,11 @@ class AlgorithmsAPI(APIBase):
     base_path = "algorithms/"
 
 
-class ResultsAPI(APIBase):
+class AlgorithmResultsAPI(APIBase):
     base_path = "algorithms/results/"
 
 
-class JobsAPI(APIBase):
+class AlgorithmJobsAPI(APIBase):
     base_path = "algorithms/jobs/"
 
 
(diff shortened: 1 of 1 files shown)


gcapi/gcapi.py
```diff
@@ -34,6 +34,18 @@ class WorkstationSessionsAPI(APIBase):
     base_path = "workstations/sessions/"
 
 
+class AlgorithmsAPI(APIBase):
+    base_path = "algorithms/"
+
+
+class ResultsAPI(APIBase):
```
jmsmkn: Probably should be `AlgorithmResultsAPI` as we also have evaluation results

gcapi/gcapi.py
```diff
@@ -34,6 +34,18 @@ class WorkstationSessionsAPI(APIBase):
...
+    base_path = "algorithms/results/"
+
+
+class JobsAPI(APIBase):
```
jmsmkn: Probably should be AlgorithmJobsAPI as we also have evaluation jobs

gcapi/gcapi.py
```diff
@@ -57,6 +69,9 @@ def __init__(
         self.images = ImagesAPI(client=self)
         self.reader_studies = ReaderStudiesAPI(client=self)
         self.sessions = WorkstationSessionsAPI(client=self)
+        self.algorithms = AlgorithmsAPI(client=self)
+        self.results = ResultsAPI(client=self)
```
jmsmkn: Probably should be `self.algorithm_results` as we also have evaluation results

gcapi/gcapi.py
```diff
@@ -57,6 +69,9 @@ def __init__(
...
+        self.jobs = JobsAPI(client=self)
```
jmsmkn: Probably should be `self.algorithm_jobs` as we also have evaluation jobs

PR #6, opened 2019-08-06
JSON verification for API requests

pkcakeout:
(messed up the branch-name)

- Implements jsonschema verification on gcapi-requests to verify that the API did not change.


gcapi/gcapi.py
```diff
@@ -21,41 +26,80 @@ def is_uuid(s):
...
+            os.path.dirname(os.path.abspath(__file__)), "schemas", filename
+        )
+
+        try:
+            with open(filename, "r") as f:
+                jsn = json.load(f)
+            return jsonschema.Draft7Validator(jsn)
+        except ValueError as e:
+            print("Error parsing '{}': {}".format(e, filename), file=sys.stderr)
+            return None
```
jmsmkn: Should this function return `None` here? Wouldn't it be better to raise an error here and let the callers handle it?

gcapi/gcapi.py
```diff
@@ -148,7 +192,9 @@ def _validate_url(self, url):
         if not url.startswith(self._base_url):
             raise RuntimeError("{} does not start with {}".format(url, self._base_url))
 
-    def __call__(self, method="GET", url="", path="", params=None, json=None, extra_headers={}):
+    def __call__(
+        self, method="GET", url="", path="", params=None, json=None, extra_headers={}
```
jmsmkn: `extra_headers` is mutable!

gcapi/gcapi.py
```diff
@@ -67,7 +111,8 @@ def send(self, **kwargs):
             method="POST",
             path=self.base_path,
             json=kwargs,
-            extra_headers={"Content-Type": "application/json"})
+            extra_headers={"Content-Type": "application/json"},
```
jmsmkn: Shouldn't every request have this?

Pipfile
```diff
@@ -0,0 +1,12 @@
+[[source]]
```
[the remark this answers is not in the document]
pkcakeout: Ha! Did not know about the `-e .` trick. Just looked it up - yeah we can remove the `Pipenv` files then!

PR #8, opened 2019-08-22
Futurized for MeVisLab upgrade

miriam-groeneveld:


jmsmkn:
This PR removes a few recent developments, I think that this is because the file was copied from an old submodule checkout. Would you be able to make a fresh clone of gcapi (or pull from master), make a new branch, run both stages of futureise, and make a new PR with the result?

miriam-groeneveld:
Hi James, I'm busy with that at the moment, although I don't understand
completely why I have old code, I did a git clone in a new folder
yesterday.

miriam-groeneveld:
version problems, will create a new branch

Issue #14, opened 2019-09-27
Upload files in chunks


haimasree:
Write a function which will upload files in chunk to grand challenge using rest framework.

jmsmkn:
Be sure to add a post API endpoint too, something like: https://github.com/DIAGNijmegen/rse-gcapi/blob/master/gcapi/gcapi.py#L159

PR #15, opened 2019-10-01
Upload chunked file

haimasree:
Closes: #14 
-  Added a post api endpoint for chunked-uploads.
-  Added code to make specific post request to this endpoint.
- Modified ```Client.__call__()``` to handle both "GET" and "POST". 
- Added a basic test.


gcapi/gcapi.py
```diff
@@ -243,15 +347,32 @@ def __call__(self,
             extra_headers["Content-Type"] = "application/json"
 
         self._validate_url(url)
+        if method == "GET":
```
jmsmkn: This should not be done with an if statement, you should be able to achieve this with variable unpacking.

gcapi/gcapi.py
```diff
@@ -25,14 +28,25 @@ def is_uuid(s):
...
+)
+
+
+def load_input_data(input_file):
+    with open(os.path.join(os.path.dirname(__file__), input_file), "rb") as f:
+        return f.read()
 
 
+def generate_new_upload_id(content):
+    return f"{hash(content)}_{time()}_{random()}"
```
jmsmkn: We have to support Python 2.7 still so cannot use f-strings :-(

gcapi/gcapi.py
```diff
@@ -197,10 +211,95 @@ class AlgorithmJobsAPI(APIBase):
...
+        ----------
+        file_info: dict
+        Contains information about the chunk, start_byte, end_byte and upload_id.
+
+        Raises
+        ------
+        ConnectionError
+            Raised if the chunk cannot be uploaded within 3 attempts.
+        """
+        filename = "whatever.bin"
```
jmsmkn: Last thing - the filename should come from the filename that the user has chosen to upload

PR #17, opened 2019-10-11
Add test server

haimasree:


jmsmkn:
I'm going to close this PR as I think that the best course of action is to start again by branching from master to keep the git history clean.

PR #18, opened 2019-10-14
Adds test against a local grand challenge server

### LINKED ISSUE #16: Add tests against a development server
- In the tests, add a checkout for `comic/grand-challenge.org`, make sure that it is not in the gcapi directory
- Run `cycle_docker_compose`
- Run the tests using the users in `init_gc_demo`, extend those if needed

jmsmkn:
To run this locally, ensure that grand challenge is running.

Closes #16

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,4 @@
+[submodule "tests/grand-challenge.org"]
+	path = tests/grand-challenge.org
+	url = https://github.com/comic/grand-challenge.org
+	branch = master
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,4 +1,4 @@
-[submodule "tests/grand-challenge.org"]
-	path = tests/grand-challenge.org
+[submodule "grand-challenge.org"]
+	path = grand-challenge.org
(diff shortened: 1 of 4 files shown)


jmsmkn:
Ok, please add that.

I'll reply on here:

```
addons:
  hosts:
    - gc.localhost
```

The host that is performing the request does not know how to resolve the hostname `gc.localhost`. We can give the host a map of hostnames to ip addresses in /etc/hosts. This block sets up /etc/hosts in travis to map the hostname `gc.localhost` to `127.0.0.1`. 

```
before_install:
    - sudo service postgresql stop
    - sudo service redis-server stop
```

Travis includes a postgresql server and redis server on their default ports to help people with testing. We need to stop those servers as the grand challenge docker compose file also tries to start it's own redis and postgres servers on the same ports. If we do not disable the travis servers, the ones used by grand challenge cannot start.

```
norecursedirs = grand-challenge.org
```

This tells tox not to look in the grand-challenge.org folder for tests, as we only want to find the tests for gcapi (the grand-challenge.org tests we know are already running correctly as we have checked out a version of master).

PR #21, opened 2019-12-04
Move duplicate definitions into separate file

### LINKED ISSUE #20: Merge several reused definitions into one file
I noticed several duplicate definitions in the JSON schemas, for example `uuid`, `api-link`, `gc-db-object`, `line-object`, etc. I think this can be refactored by moving the duplicates to a separate `definitions.json` file and reference them in the schemas using: 
`{ "$ref": "definitions.json#/name" }`
Is there a reason this is not used currently @jmsmkn / @pkcakeout?
If not, I can do this while I'm creating a JSON schema for `LandmarkAnnotationSet` objects since I also need to reuse several of those definitions.

HarmvZ:
Closes #20

### DIFF
--- a/gcapi/schemas/answer.json
+++ b/gcapi/schemas/answer.json
@@ -1,135 +1,16 @@
 {
     "$schema": "http://json-schema.org/draft-07/schema#",
-    "definitions": {
-        "uuid": {
-            "type": "string",
-            "pattern": "^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$"
-        },
-        "api-link": {
-            "type": "string",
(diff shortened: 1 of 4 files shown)


pkcakeout:
Did we verify that this works as expected? From what I see, the jsonschema loader does not (yet) support cross-referencing of schemas. Check these lines:

https://github.com/DIAGNijmegen/rse-gcapi/blob/e0ce6f285295befc23367450289e91911fdd7dd4/gcapi/gcapi.py#L81-L83

The validator has no idea where the the original data comes from since it only gets fed with parsed json. We should verify that this correctly fails for inputs that violate definitions from the `definitions.json`. Also regarindg your question @HarmvZ  in issue #20 - this is why this was not done yet. (I will post some more background information there as well)

jmsmkn:
We decided not to merge this as a more complete solution for all the repos is needed.

PR #43, opened 2020-03-17
Add retina polygon client functionality

### LINKED ISSUE #42: Add support for PolygonAnnotation
Add support for new annotation endpoints which are added in https://github.com/comic/grand-challenge.org/pull/1210

HarmvZ:
Requires changes in https://github.com/comic/grand-challenge.org/pull/1210 to be merged before tests pass.
Closes #42

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -3,12 +3,11 @@ History
 =======
 
 
-0.2.5 (2020-04-10)
+0.2.5 (unreleased)
 ------------------
 
 * Added support for retina polygon annotation sets and retina single polygon annotations
 
-=======
 0.2.4 (2020-04-03)
 ------------------
 
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -2,6 +2,11 @@
 History
 =======
 
+0.2.3 (2020-03-25)
+------------------
+
+* Added support for retina polygon annotation sets and retina single polygon annotations
+
 0.2.2 (2020-03-24)
 ------------------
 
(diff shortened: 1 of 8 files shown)


jmsmkn:
Re-running tests

HISTORY.rst
```diff
@@ -2,6 +2,13 @@
 History
 =======
 
+
+0.2.5 (2020-04-10)
+------------------
+
+* Added support for retina polygon annotation sets and retina single polygon annotations
+
+=======
```
jmsmkn, line 11:
```suggestion

0.2.5 (unreleased)
------------------

* Added support for retina polygon annotation sets and retina single polygon annotations

```

jmsmkn: Good to add to the changelog, but we have a separate release process.

gcapi/gcapi.py
```diff
@@ -297,6 +297,28 @@ def for_image(self, pk):
...
+    json_schema = import_json_schema("polygon-annotation.json")
+    modify_json_schema = import_json_schema("post-polygon-annotation.json")
+
+    def for_image(self, pk):
+        result = self._client(
+            method="GET", path=self.base_path, params={"image": pk}
+        )
+        for i in result:
+            self._verify_against_schema(i)
+        return result
```
jmsmkn, line 311: This won't work if there is more than 1 page, right? Why have a specific method for this rather than letting the user provide the params?

HarmvZ: Pagination is disabled on grand challenge for this endpoint or is that not what you mean?
Anyway, the method is indeed not necessary I can just pass the params in cirrus-core to the `list()` method, so I removed it.

jmsmkn: Pagination should be enabled by default, but if you make a get request to a list you will only ever get 100 results, then you will need to iterate over `next=`.

PR #52, opened 2020-04-30
Implement ground_truth getter through APIBase

### LINKED ISSUE #49: reader_studies.ground_truth is implemented weirdly
For all other APIs we use the sub_apis feature except for ground_truth:

https://github.com/DIAGNijmegen/rse-gcapi/blob/15d1d7e6b9d119588e531d60f86f19fda67fad7b/gcapi/gcapi.py#L262-L267

This makes the invocation to APIs in gcapi inhomogeneneous. We should use the standard sub-api feature. What is done in this function seems to be 100% covered by the `detail(pk)` function of APIBase.

pkcakeout:
* Add a LazyDict type that behaves like a dictionary but defers loading
  of the data to a later moment. The advantage here is that this allows us
  to attach extra functions to the detail view that might not require getting
  the actual object.
* Add the ground_truth detail view to the reader-study detail view
  that allows to query ground_truth data for a reader study
* Deprecate old method which bypassed the APIBase way of interactin
  with grand-challenge. I left it in there for now with a print-warning
  which we can remove later on.

Fixes #49

-------------------

This will allow us do consolidate the cirrus code:

~~~~~.py
        rs = gc_client.reader_studies
        [...]
        ground_truth = rs.ground_truth(self.study_id, pk)
~~~~~

to 

~~~~~.py
        ground_truth = self.reader_study.ground_truth.detail(pk)
~~~~~

Also this can serve as an example how this can be implemented for other recursive detail views in the future. We should likely flesh this out a bit more then.

### DIFF
--- a/gcapi/gcapi.py
+++ b/gcapi/gcapi.py
@@ -1,6 +1,7 @@
 import itertools
 import os
 import uuid
+import warnings
 from collections import UserDict
 from io import BytesIO
 from json import load
@@ -286,10 +287,11 @@ class ReaderStudiesAPI(APIBase):
         return result
 
     def ground_truth(self, pk, case_pk):
-        print(
-            "GCAPI DEPRECATION WARNING: Please use "
-            "reader_studies.detail(pk).ground_truth.detail(case_pk) instead of "
-            "reader_studies.ground_truth(self, pk, case_pk)"
+        warnings.warn(
+            DeprecationWarning(
+                "Please use reader_studies.detail(pk).ground_truth.detail(case_pk) "
+                "instead of reader_studies.ground_truth(self, pk, case_pk)"
+            )
         )
         return self.detail(pk).ground_truth.detail(case_pk)
 
--- a/gcapi/gcapi.py
+++ b/gcapi/gcapi.py
@@ -1,6 +1,7 @@
 import itertools
 import os
 import uuid
+from collections import UserDict
 from io import BytesIO
 from json import load
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


gcapi/gcapi.py
```diff
@@ -263,12 +280,27 @@ class ReaderStudiesAPI(APIBase):
...
+    def detail(self, pk):
+        result = LazyDict(lambda: super().detail(pk))
+        result.ground_truth = GroundTruthAPI(self._client, pk)
+        return result
+
     def ground_truth(self, pk, case_pk):
-        result = self._client(
-            method="GET",
-            path=urljoin(self.base_path, pk + "/ground-truth/" + case_pk),
+        print(
```
jmsmkn, line 289: Don't print warnings! Make this a warning.

gcapi/gcapi.py
```diff
@@ -263,12 +280,27 @@ class ReaderStudiesAPI(APIBase):
     answers = None  # type: ReaderStudyAnswersAPI
     questions = None  # type: ReaderStudyQuestionsAPI
 
+    def detail(self, pk):
+        result = LazyDict(lambda: super().detail(pk))
```
jmsmkn, line 284: So now we have to call `result.data` on everything before we can use it?

pkcakeout: No... the `.data` think is part of the `UserDict` contract. You could call it but you can just use the `LazyDict` directly as well. It _is_ a dict. The [-> doc](https://docs.python.org/3/library/collections.html?highlight=userdict#collections.UserDict) is really short.

pkcakeout: I tested this against cirrus-core already btw. It is used over there. It just works.

gcapi/gcapi.py
pkcakeout: Yeah, I was wondering about this. Any ideas or preferences on how to solve this. `logging.warning()` ?

jmsmkn: https://docs.python.org/3/library/exceptions.html#DeprecationWarning

pkcakeout: Cool thanks! Fixed!

gcapi/gcapi.py
pkcakeout: Added a small test case to illustrate how lazy-dict is used.

PR #53, opened 2020-04-30
Add job filter endpoint for corads plugins

pkcakeout:


jmsmkn:
What is going on with this PR? None of this makes any sense to me, why is any of this needed?

pkcakeout:
Hi James!

Yeah, I did not get around commenting this because getting the tooling right on my machine cost quite a bit of time and I though I write down the rationale today. Most of the pr is this function `parse_uuid_from_url` that I decorated every single function that takes a `pk` with. The reason why I added it is that there are several locations in cirrus-core where people (me or Mike or who ever)  have written hand-written code to parse the pk out of a url. I can name two instances from the top of my head:

* https://github.com/DIAGNijmegen/rse-cirrus-core/blob/a7d60fc57f2e10acb82b2ffaeee957876d9f8a12/src/cirrus/DIAG/CIRRUSCore/Modules/Macros/CIRRUSCoreWeb/CIRRUSCoreWeb.py#L1450

* https://github.com/DIAGNijmegen/rse-cirrus-core/blob/a7d60fc57f2e10acb82b2ffaeee957876d9f8a12/src/cirrus/DIAG/CIRRUSCore/Modules/Macros/CIRRUSCoreWeb/CIRRUSCoreWeb.py#L2207-L2211

I remember to have seen this in the past as well. However, it appears to be come a more common necessity. You might say that we can just feed the urls into the `client.get` functions, but this will only bring us so far as the pks are required to be used as query parameters like in the new `def by_input_image(self, pk):` function. Instead of doing this as a one-off I annotated every function that can ingest a pk with the parser to allow easier interactions with the API in a homogeneous way.

**Feature:** So what we can do now is to supply api_urls as pk for every function which reduces the 
need for ad-hoc hacks like the ones I linked above.

pkcakeout:
Also I should mention that I only wrote all of this because it will be directly handy to have for the CORADS plugin. I would have written something like this for that otherwise, so I just put it in here so that it can be usefule for everybody.

gcapi/gcapi.py
```diff
@@ -282,12 +379,17 @@ class AlgorithmResultsAPI(APIBase):
 class AlgorithmJobsAPI(APIBase):
     base_path = "algorithms/jobs/"
 
+    @filter_arguments(None, parse_uuid_from_url)
+    def by_input_image(self, pk):
+        return self.iterate_all(params={"image": pk})
```
jmsmkn, line 384: Wait, all this complexity is here just so you don't need to do `image.pk`? The `parse_uuid_from_url` is fragile and too much magic. The alternative would be on the grand challenge side to accept filters by api urls rather than pks, but not this.

jmsmkn:
Yes, then those things need to be refactored. You can get the pks from the objects that are returned. It is super dangerous to try and guess the pk from the url, we do not want to generalise that hack.

gcapi/gcapi.py
pkcakeout: I just pretty much replicated what is already happening on the cirrus-core side. Ii made sense from my side to embed this in here instead.

`image.pk` is a bit too simple of an example though as there are many places where gcapi does return api_urls but no pks. Example: https://grand-challenge.org/api/v1/algorithms/e5d03359-4c31-46de-9dd6-e077d925dc4d/ - the jobs specifically do, too, but I do not have an example url handy.

However, if grand-challenge could ingest api_urls as alternative for pks that would be great, then we would not need this and can clean up the code in cirrus-core anyway! Certainly an alternative!

pkcakeout:
> Yes, then those things need to be refactored. You can get the pks from the objects that are returned. It is super dangerous to try and guess the pk from the url, we do not want to generalise that hack.

Like I intended to say: this is not possible in all locations at the moment because objects tend to refer to other objects by api_url. To prevent us from having to make that extra get-call, just to obtain the pk for an object, cirrus is parsing it out of the url. That seems to be a common usecase. "Just get the PR of *that thing* referenced in this struct by *api_url* and put it in *that other query* "

pkcakeout:
Hey wait! I can remove the extra stuff, but I still need the `def by_input_image(self, pk):` bit!

jmsmkn:
It is trivial to enable filtering by api url too on the grand challenge
side, lets do that instead:
https://django-filter.readthedocs.io/en/master/guide/rest_framework.html#using-the-filterset-fields-shortcut

jmsmkn:
Issue for API extension here: https://github.com/comic/grand-challenge.org/issues/1347

PR #57, opened 2020-05-18
Replace generic "modification" json schema with a request specific dict

pkcakeout:
* Unifies GET/POST/... request validations through use of a lookup table
* Removed the need for validate=False feature - just do not include a validation schema
  for the request type that should not be validated.

### DIFF
--- a/gcapi/gcapi.py
+++ b/gcapi/gcapi.py
@@ -97,9 +97,12 @@ class APIBase:
     base_path = ""
     sub_apis: Dict[str, Type["APIBase"]] = {}
 
-    json_schema = None
+    validation_schemas = None  # type: Dict[str, jsonschema.Draft7Validator]
 
     def __init__(self, client):
+        if self.validation_schemas is None:
+            self.validation_schemas = {}
+
         if isinstance(self, ModifiableMixin):
             ModifiableMixin.__init__(self)
 
@@ -109,8 +112,9 @@ class APIBase:
             setattr(self, k, api(self._client))
 
     def _verify_against_schema(self, value):
-        if self.json_schema is not None:
-            self.json_schema.validate(value)
+        schema = self.validation_schemas.get("GET")
+        if schema is not None:
+            schema.validate(value)
 
     def list(self, params=None):
         result = self._client(method="GET", path=self.base_path, params=params)
@@ -154,19 +158,15 @@ class APIBase:
 class ModifiableMixin:
     _client = None  # type: Client
 
-    modify_json_schema = None  # type: jsonschema.Draft7Validator
-
     def __init__(self):
         pass
 
-    def _process_post_arguments(self, post_args):
-        if self.modify_json_schema is not None:
-            self.modify_json_schema.validate(post_args)
-
-    def _validate_data(self, data):
+    def _process_request_arguments(self, method, data):
         if data is None:
             data = {}
-        self._process_post_arguments(data)
+        schema = self.validation_schemas.get(method)
+        if schema:
+            schema.validate(data)
         return data
 
     def _execute_request(self, method, data, pk):
@@ -177,9 +177,8 @@ class ModifiableMixin:
         )
         return self._client(method=method, path=url, json=data)
 
-    def perform_request(self, method, data=None, pk=False, validate=True):
-        if validate:
-            data = self._validate_data(data)
+    def perform_request(self, method, data=None, pk=False):
+        data = self._process_request_arguments(method, data)
         return self._execute_request(method, data, pk)
 
     def send(self, **kwargs):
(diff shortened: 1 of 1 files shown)


jmsmkn:
Why are we doing post or patch validations here anyway? The backend does far more robust validation (not only schema, but values too), and it's a complete pain to keep these up to date in all places.

pkcakeout:
Well. It is legacy and helps testing locally. It is the old problem of agreeing to the same _interface_. And the interface has a client-side and a server-side. This is the client-side of the interface. Nicely specced in jsonschema.

That at least was our original intent here. I know that it is a pain but I also did not want to break any functionality. So instead of stripping gcapi bare of any validation, I just made it more specific to the cases where it should be used. We should discus in another ticket if we want to remove it completely. I am a bit on the fence about that.

gcapi/gcapi.py
```diff
@@ -109,8 +112,9 @@ def __init__(self, client):
             setattr(self, k, api(self._client))
 
     def _verify_against_schema(self, value):
-        if self.json_schema is not None:
-            self.json_schema.validate(value)
+        schema = self.validation_schemas.get("GET")
+        if schema is not None:
+            schema.validate(value)
```
jmsmkn, line 117: Will all these schemas work if the schema is `None`, as in, a nullable field? I guess the schema itself is never null, only the attributes are.

gcapi/gcapi.py
```diff
@@ -154,19 +158,15 @@ def detail(self, pk):
...
-        if self.modify_json_schema is not None:
-            self.modify_json_schema.validate(post_args)
-
-    def _validate_data(self, data):
+    def _process_request_arguments(self, method, data):
         if data is None:
             data = {}
-        self._process_post_arguments(data)
+        schema = self.validation_schemas.get(method)
+        if schema:
```
jmsmkn, line 168: `if schema is None`? What if the schema is `{}`, is this possible?

gcapi/gcapi.py
```diff
@@ -154,19 +158,15 @@ def detail(self, pk):
...
+            schema.validate(data)
```
jmsmkn, line 169: Duplicate of `_verify_against_schema`?

gcapi/gcapi.py
pkcakeout: A schema has to be a Draft7Validator instance that contains the schema. The schema was already adapted for that, see for example here:

https://github.com/DIAGNijmegen/rse-gcapi/blob/b00fb9f849fc8a39d6a90ff701addc7f2ea1f917/gcapi/schemas/answer.json#L219-L234

I only made sure that these schemas are only applied to certain types of requests - POST in this case. I did not write a PATCH jsonschema. So those go out unchecked right now.

gcapi/gcapi.py
pkcakeout: Like I wrote above. It is either not in the dict or a Draft7Validator instances. It is never a bare json dict. I think I even added the type annotation!

gcapi/gcapi.py
pkcakeout: yes, but `_verify_against_schema` only deals with GET requests as inherited from the simpler base class. This function is different because it deals with all sorts of requests and is part of the `ModifiableMixin` interface. Since this is only a Mixin I cannot use `_verify_against_schema` directly because that function is not part of the class hierarchy and vice versa. Well, I could, but I think that is very messy at that point. Also, the `_process_request_arguments` function is overridden in some implementations to offer special functionality for POST/PATCH requests only. 

So this needs to remain separate somehow. Now that I refactored it is kind of obvious that both function are very similar and likely can be joined together somehow. I don't know. Make that call. Should I unify the two for just three lines of code? Probably is more code to implement that.

jmsmkn: Urgh, that Mixin stuff is a mess, instantiated incorrectly in all places (should be mixin first, then base class). Then you could use the base method. Is all of this covered by tests? None have been added in this PR. For some reason there is no report on codecov.

pkcakeout: The order did not really matter back in the days of python2. We still use python2 initialization so whatever. The entire Mixin stuff is used extensively  and covered by tests, yes. Actually we rely on it working, the integration tests are using it, for example. I did not change anything functionally when it comes to the feature set that is currently used just implemented if in a different way.  Example test:

https://github.com/DIAGNijmegen/rse-gcapi/blob/b00fb9f849fc8a39d6a90ff701addc7f2ea1f917/tests/integration_tests.py#L43-L44

---------------------

We do *not* test any PATCH requests however in our API as far as I can see. That require an awkward programmatical initialization of a reader-study, submitting some example answers and the PATCHing them. Quite messy to do as integration tests. A requests-mock based test might work however.But we do not have a requests-mock test in gcapi yet, so I would suggest splitting this out into another issue.

jmsmkn: I mean tests of the validation and what you have worked on in this PR is working, I cannot see any.

pkcakeout: That is pretty true. Like I said: Not PATCH request tests yet. I will see what I can do.

PR #73, opened 2021-02-28
Normalise token header

jmsmkn:
We currently support two methods of specifying the authentication token, through arguments and through the environment variable, and two formats for the token `token` and `bearer`. This standardises all methods of setting the token, and will always produce a `bearer` token. 

`bearer` support is currently missing from the retina endpoints, but will be standardised in https://github.com/comic/grand-challenge.org/pull/1740. Updating CIRRUS is not required as it will pass the token without modification right now. This should be released to pypi alongside 1740.

### DIFF
--- a/gcapi/gcapi.py
+++ b/gcapi/gcapi.py
@@ -1,5 +1,6 @@
 import itertools
 import os
+import re
 import uuid
 from io import BytesIO
 from json import load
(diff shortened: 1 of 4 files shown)


HarmvZ (approved):
LGTM, but this also needs a minor/patch version bump so CIRRUS can use the new version, right?

jmsmkn:
Will do as part of a separate release process.

PR #76, opened 2021-05-20
test create job

miriam-groeneveld:
New function run_external_job. This function can create a job for algorithms with multiple inputs. There are a few options:
1. Add inputs for interfaces with image kind, 2 sub options:
1a: Uploading files to create images. In this case a list of files should be provided. An upload session is started for each input. The upload_session_pk is added to the input.
1b: Adding existing images. In this case the hyperlink to the image should be provided.
2: Add inputs for other kinds (int, string, annotations etc. In that case the value should be provided.

### DIFF
--- a/gcapi/gcapi.py
+++ b/gcapi/gcapi.py
@@ -128,7 +128,7 @@ class APIBase:
         params["offset"] = offset
         params["limit"] = limit
         result = self._client(
-            method="GET", path=self.base_path, params=params,
+            method="GET", path=self.base_path, params=params
         )["results"]
         for i in result:
             self._verify_against_schema(i)
@@ -252,9 +252,7 @@ class ReaderStudyAnswersAPI(APIBase, ModifiableMixin):
 
 class ReaderStudiesAPI(APIBase):
     base_path = "reader-studies/"
-    validation_schemas = {
-        "GET": import_json_schema("reader-study.json"),
-    }
+    validation_schemas = {"GET": import_json_schema("reader-study.json")}
 
     sub_apis = {
         "answers": ReaderStudyAnswersAPI,
@@ -280,7 +278,7 @@ class AlgorithmResultsAPI(APIBase):
     base_path = "algorithms/results/"
 
 
-class AlgorithmJobsAPI(APIBase):
+class AlgorithmJobsAPI(APIBase, ModifiableMixin):
     base_path = "algorithms/jobs/"
 
     def by_input_image(self, pk):
@@ -539,6 +537,39 @@ class Client(Session):
         else:
             return response
 
+    def run_external_job(
+        self, *, algorithm_name: str = None, inputs: Dict[str, any]
+    ):
+        algorithm = self._get_algorithm(algorithm_name=algorithm_name)
+        input_interfaces = {ci["title"]: ci for ci in algorithm["inputs"]}
+
+        for ci in input_interfaces:
+            if ci not in inputs and input_interfaces[ci]["default_value"] is None:
+                raise ValueError(f"{ci} is not provided")
+
+        job = {"algorithm_title": algorithm_name, "inputs": []}
+        for input_title, value in inputs.items():
+            ci = input_interfaces.get(input_title, None)
+            if not ci:
+                raise ValueError(
+                    f"{input_title} is not an input interface for this algorithm"
+                )
+
+            i = {"interface_title": input_title}
+            if ci["kind"] in ("Image", "Segmentation", "Heatmap"):
+                if isinstance(value, list):
+                    raw_image_upload_session = self.upload_cases(
+                        files=value, process=False
+                    )
+                    i["upload_pk"] = raw_image_upload_session["pk"]
+                elif isinstance(value, str):
+                    i["image"] = value
+            else:
+                i["value"] = value
+            job["inputs"].append(i)
+
+        return self.algorithm_jobs.create(**job)
+
     def upload_cases(
         self,
         *,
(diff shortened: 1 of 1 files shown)


gcapi/gcapi.py
```diff
@@ -578,6 +610,10 @@ def upload_cases(
             The slug of the archive to use.
         reader_study
             The slug of the reader study to use.
+        process
+            Should the images be processed (start a job for the algorithm or add to
+            archive or reader study). Set to False when using an algorithm with
+            multiple inputs.
```
jmsmkn, line 616: I think we can remove this kwarg. Don't give the users the option, we can remove the legacy endpoint. If the have multiple images they want to run then they can loop around this function. Make processing far more efficient.

gcapi/gcapi.py
```diff
@@ -539,13 +537,47 @@ def __call__(
         else:
             return response
 
+    def run_external_job(
+        self, *, algorithm_name: str = None, inputs: Dict[str, any]
```
jmsmkn, line 541: For API consistency `algorithm_name` -> `algorithm` and take a slug (see upload_cases). Then, we can refactor `_get_algorithm` too.

gcapi/gcapi.py
```diff
@@ -539,13 +537,47 @@ def __call__(
...
+    ):
+        algorithm = self._get_algorithm(algorithm_name=algorithm_name)
+        input_interfaces = {ci["title"]: ci for ci in algorithm["inputs"]}
```
jmsmkn, line 544: Don't use titles, they are not unique, use slugs instead.

gcapi/gcapi.py
```diff
@@ -539,13 +537,47 @@ def __call__(
...
+        job = {"algorithm_slug": algorithm["slug"], "inputs": []}
+        for input_title, value in inputs.items():
+            ci = input_interfaces.get(input_title, None)
+            if not ci:
+                raise ValueError(
+                    f"{input_title} is not an input interface for this algorithm"
+                )
+
+            i = {"interface": ci["slug"]}
+            if ci["kind"] in ("Image", "Segmentation", "Heatmap"):
```
jmsmkn, line 559: Maybe we need to expose whether this takes an image in the API as this could get stale.

gcapi/gcapi.py
```diff
@@ -618,8 +654,24 @@ def upload_cases(
...
 
         return raw_image_upload_session
+
+    def _get_algorithm(self, algorithm_name: str) -> dict:
+        """Get the algorithm for the given algorithm name. """
+        algorithms = [
+            a
+            for a in list(self.algorithms.list()["results"])
+            if a["title"] == algorithm_name
+        ]
```
jmsmkn, line 670: `algorithm_name` -> `algorithm`, and add a filter as this will fetch all algorithms the user has access to from the API, rather than 1. You want `/api/v1/algorithms/?slug=blah`, should return 1 result.

tests/conftest.py
```diff
@@ -11,15 +11,15 @@
 from tests.integration_tests import ADMIN_TOKEN
 
 
-@pytest.yield_fixture(scope="session")
+@pytest.fixture(scope="session")
```
jmsmkn, line 14: Is it really correct to change here? We are yielding and tearing down in tests.

gcapi/gcapi.py
```diff
@@ -539,13 +537,47 @@ def __call__(
...
+        self, *, algorithm_name: str = None, inputs: Dict[str, any]
+    ):
+        algorithm = self._get_algorithm(algorithm_name=algorithm_name)
+        input_interfaces = {ci["title"]: ci for ci in algorithm["inputs"]}
+
+        for ci in input_interfaces:
+            if ci not in inputs and input_interfaces[ci]["default_value"] is None:
+                raise ValueError(f"{ci} is not provided")
+
+        job = {"algorithm_slug": algorithm["slug"], "inputs": []}
```
jmsmkn, line 550: Should we send `algorithm_slug`, or change the serializer on the gc side an call it `algorithm` and then use a pk or hyperlinked related field?

tests/conftest.py
miriam-groeneveld: The yield_fixture is deprecated according to the docs and does the same as fixture:
https://docs.pytest.org/en/latest/yieldfixture.html

jmsmkn: Nice, didn't know that, thanks!

gcapi/gcapi.py
miriam-groeneveld: sorry, not sure what you mean :).

gcapi/gcapi.py
miriam-groeneveld: Ok, changed it to HyperlinkedRelatedField.

gcapi/gcapi.py
miriam-groeneveld: SuperKind sounds very friendly :+1: , most inline with description of kind as well.

gcapi/gcapi.py
```diff
@@ -539,6 +538,123 @@ def __call__(
...
+        for file_id, filename in uploaded_files.items():
+            self.raw_image_upload_session_files.create(
+                upload_session=raw_image_upload_session["api_url"],
+                staged_file_id=file_id,
+                filename=filename,
+            )
+
+        return raw_image_upload_session
+
+    def run_external_job(self, *, algorithm: str, inputs: Dict[str, Any]):
```
jmsmkn, line 576: `Any` is probably too broad, maybe `Union[List[Path],str]` (where the str is actually a URL to an image? or upload session? or a pk?), also should have a return type of algorithm job. Not worth fixing as types right now are a huge rabbit hole for little gain here.

jmsmkn (approved):
LGTM!

PR #77, opened 2021-05-29
Fix CI

jmsmkn:
The way fixtures are generated for grand challenge was changed in https://github.com/comic/grand-challenge.org/pull/1851, we need to grab a couple of extra files here for the make command to work now. Also removes some xfail markers for tests which were passing after a PR merge in grand challenge and fixes the reader study schema as the hanging list has always been able to be null too.

#76 should be rebased on to this once on master.


miriam-groeneveld (approved):
LGTM

PR #82, opened 2021-08-27
Switch to httpx client

pkcakeout:
Fixed the file uploads. Took a while to debug, but the file upload interface in httpx is different from requests:

https://github.com/encode/httpx/blob/437b55c5202a034e8e5fdce4be1b96198d33be31/docs/quickstart.md#sending-multipart-file-uploads

### DIFF
--- a/gcapi/gcapi.py
+++ b/gcapi/gcapi.py
@@ -8,7 +8,6 @@ from random import randint, random
 from time import sleep, time
 from typing import Any, Dict, List, Type
 from urllib.parse import urljoin, urlparse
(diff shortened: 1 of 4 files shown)


jmsmkn (approved):
LGTM, we should get this on master and cut another alpha

pkcakeout:
> LGTM, we should get this on master and cut another alpha

That is brave :smile:  I did not get around to do the improvements you suggested...

jmsmkn:
That's ok, that is what alpha releases are for!

PR #86, opened 2021-10-05
Add params to detail endpoint

### LINKED ISSUE #85: Add helper function to retrieve item details using slug
There is now the `def get_algorithm(self, algorithm: str) -> dict` function that returns the details of an algorithm given its slug. This would be very helpful for other items as well, e.g., archives. Maybe this could be integrated into the `.detail()` function, which currently only accepts IDs/PKs.

jmsmkn:
Closes #85

### DIFF
--- a/gcapi/gcapi.py
+++ b/gcapi/gcapi.py
@@ -13,6 +13,8 @@ import jsonschema
 from httpx import Client as SyncClient
 from httpx import HTTPError, Timeout
 
+from .exceptions import MultipleObjectsReturned, ObjectNotFound
+
 
 def is_uuid(s):
     try:
@@ -147,11 +149,24 @@ class APIBase:
             yield from current_list
             offset += req_count
 
-    def detail(self, pk):
-        result = self._client(
-            method="GET", path=urljoin(self.base_path, pk + "/")
-        )
-        self._verify_against_schema(result)
+    def detail(self, pk=None, **kwargs):
+        if len([x for x in [pk, kwargs] if x]) != 1:
+            raise ValueError("Only one of pk or kwargs must be specified")
+
+        if pk is not None:
+            result = self._client(
+                method="GET", path=urljoin(self.base_path, pk + "/")
+            )
+            self._verify_against_schema(result)
+        else:
+            results = [v for v in self.page(params=kwargs)]
+            if len(results) == 1:
+                result = results[0]
+            elif len(results) == 0:
+                raise ObjectNotFound
+            else:
+                raise MultipleObjectsReturned
+
         return result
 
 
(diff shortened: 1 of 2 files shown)


gcapi/gcapi.py
```diff
@@ -147,11 +149,24 @@ def iterate_all(self, params=None):
...
-        self._verify_against_schema(result)
+    def detail(self, pk=None, **params):
+        if len([x for x in [pk, params] if x]) != 1:
+            raise ValueError("Only one of pk or params must be specified")
+
+        if pk is not None:
+            result = self._client(
+                method="GET", path=urljoin(self.base_path, pk + "/")
+            )
+            self._verify_against_schema(result)
```
nlessmann, line 160: Is there a reason that this verification is done only for "by pk" requests?

nlessmann (commented):
Nice, looks great!

gcapi/gcapi.py
jmsmkn: It's being done in the `page` method https://github.com/DIAGNijmegen/rse-gcapi/pull/86/files#diff-0b0fdcdf902a8e6c2a49092303d186e4ef201b468ff23b4e0ca1446218c6b9b7R128-R137

gcapi/gcapi.py
```diff
@@ -147,11 +149,31 @@ def iterate_all(self, params=None):
...
             offset += req_count
 
-    def detail(self, pk):
-        result = self._client(
-            method="GET", path=urljoin(self.base_path, pk + "/")
-        )
-        self._verify_against_schema(result)
+    def detail(self, pk=None, **params):
+        if len([x for x in [pk, params] if x]) != 1:
+            raise ValueError("Only one of pk or params must be specified")
```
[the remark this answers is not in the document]
pkcakeout: If you want to catch the "none supplied" case as well, xor would likely be "correct": `bool(a) ^ bool(b)` but then the readability is also not super-good...

tests/integration_tests.py
```diff
@@ -196,3 +197,50 @@ def test_create_job_with_upload(local_grand_challenge, files):
...
+
+    with pytest.raises(ObjectNotFound):
+        c.reader_studies.detail(**{key: "foo"})
+
+
+def test_detail_multiple_objects(local_grand_challenge):
+    c = Client(token=ADMIN_TOKEN, base_url=local_grand_challenge, verify=False)
+
+    file_to_upload = os.path.join(
+        os.path.dirname(os.path.realpath(__file__)), "testdata", "rnddata"
```
pkcakeout, line 240: Random find: I am pretty sure that abspath is to preferred over realpath in pretty much all cases. This mainly has to do with how libraries - like this snippet of code - are resolving the place where data is loaded. If someone externally decides to link this single test file into some other environment with a symbolic, realpath will resolve that symbolic link and point to the original path even though that is not the file at the location where the user invoked it.

If one uses abspath instead, you also get a canonical representation of the path that includes the symbolic links, and therefore still points in the environment that the uses was calling the file from. Python internally clearly prefers abspath as it ignores any kind of symbolic links in its library paths it seems. Node-js falls apart when there is a symbolic links in any path because library developers seem quite confused with symbolic links and like to mix abspath with realpath :P

Anyway, random find. I think this does not need to be a blocker and can be fixed when it becomes a problem.

pkcakeout (approved):

tests/integration_tests.py
jmsmkn: Yeah, that was [copy-pasted from a test above](https://twitter.com/kelseyhightower/status/1022837097743319040). Test code so not worried about it.

PR #88, opened 2021-10-14
Add support for the S3 Multipart Upload API

jmsmkn:
Pretty big change where the tests will work once https://github.com/comic/grand-challenge.org/pull/2110 lands on master. Changes to the logic:

- An upload session only needs to be created once all the files have been uploaded, no need to patch it later
- As PUT requests will be made directly to S3 the auth token is removed from the default set of headers and added back manually for grand challenge URLs
- I've added some error logging to help users decipher the error codes from grand challenge
- The implementation follows the reference https://github.com/comic/grand-challenge.org/blob/8b923456ffced78ac9f049312e787d05bd382853/app/tests/uploads_tests/test_views.py#L10-L75
- I've chosen a chunk size of 32 MB with 5 URLs being prefetched each time, both adjustable but it is a balance as to how much load we put on the grand challenge servers, and how much the user can upload in the 1 hour the URLs are valid for
- Several deprecated functions. methods and endpoints are removed
- I did not find any tests for uploading an image answer so unsure how it's done in CIRRUS

Limitations:
- create multipart upload, generate_presigned_urls, complete_multipart_upload are NOT retried, maybe they should be
- _put_chunk is only retried on status 409, 423 or 5xx, this matches the uppy implementation
- Chunks are uploaded sequentially, you could fetch the next urls and then upload those chunks in parallel in async, if you want

### DIFF
--- a/tests/integration_tests.py
+++ b/tests/integration_tests.py
@@ -245,8 +245,11 @@ def test_detail_multiple_objects(local_grand_challenge):
     with pytest.raises(MultipleObjectsReturned):
         c.uploads.detail(slug="")
 
+
 def test_auth_headers_not_sent():
     c = Client(token="foo")
-    response = c.uploads._put_chunk(chunk=BytesIO(b"123"), url="https://httpbin.org/put")
+    response = c.uploads._put_chunk(
+        chunk=BytesIO(b"123"), url="https://httpbin.org/put"
+    )
     sent_headers = response.json()["headers"]
     assert not set(c._auth_header.keys()) & set(sent_headers.keys())


gcapi/gcapi.py
```diff
@@ -353,115 +337,145 @@ class RetinaETDRSGridAnnotationsAPI(APIBase, ModifiableMixin):
...
+        return response["presigned_urls"]
+
+    def _put_chunk(self, *, chunk, url):
+        num_retries = 0
+        e = Exception
 
-            start_byte += len(chunk)
+        while num_retries < self.max_retries:
+            try:
+                result = self._client.request(
```
HarmvZ, line 441: Maybe good to add a test to check that the token is not sent along with this request. When you only see this code, the difference between `self._client(...)` and `self._client.request(...)` is unclear so I can imagine that someone switch these up by accident.

jmsmkn: Done using httpbin, not sure that this is the best method.

PR #95, opened 2021-11-02
Update HISTORY.rst with follow_requests warning

pkcakeout:
Well, it looks like many people had problems with this, so I added it to the readme. I am not sure if what I wrote is _true_ (that this preset changed) but I _think_ it did with the switch to httpx ?

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -7,6 +7,7 @@ History
 
 * **Breaking Change** Switched the backend from ``requests`` to ``httpx``
 * **Breaking Change** Removed ``client.get_algorithm(algorithm=...)``, use ``client.algorithms.detail(slug=...)`` instead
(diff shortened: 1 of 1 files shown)


jmsmkn:
Okay I see where it's coming from now. `allow_redirects` in `requests` is _only_ for POST/PUT/DELETE requests: GET requests would auto-redirect by default. In `httpx` they changed the default behaviour in 0.20.0, I agree with the reasoning, see https://github.com/encode/httpx/discussions/1785. So, it is a breaking change, but rather than users allowing redirects for 301s they should update their code to use the actual endpoints.

pkcakeout:
Should we then not still merge this as a warning that the changed backends have changed the behavior of the library?

jmsmkn:
I added a thing to the release on GitHub as PyPi won't be updated until the next release https://github.com/DIAGNijmegen/rse-gcapi/releases/tag/v0.5.0

PR #96, opened 2021-11-10
Add a method for downloading images

nlessmann:
Since downloading images and masks is a common task, I think it would be helpful to have a helper function that does that. A user might often have only the UUID of the image from the algorithm job details so that downloading the actual image requires multiple steps (query for list of files associated with the image, download the files).

I've implemented this, but I'm struggling a bit to understand the new structure with all these `yield` calls, so I'm not sure whether I've done that correctly (it works though, at least with the regular client).

This method could also be added to the Client class instead of the ImagesAPI, similar to the `upload_cases()` method.

I would appreciate some feedback regarding the two points above. Thanks!

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -2,6 +2,8 @@
 History
 =======
 
+* Added ``client.images.download``
+
 0.5.0 (2021-11-01)
 ------------------
 
--- a/README.rst
+++ b/README.rst
@@ -164,6 +164,25 @@ You can refresh the session object with
 
 and check the session status with ``session["status"]``.
 
+Downloading Files
+-----------------
+
+An image can consist of one or multiple files, such as a single mha file or a dzi and tiff file. You can download all files
+associated with an image.
+
+.. code:: python
+
+    from pathlib import Path
+
+    downloaded_files = c.images.download(pk="...", filename=Path("path/to/output"))
+
+You can also use other parameters to identify the image, such as the API URL, and you can also supply the "files" list directly.
+
+... code:: python
+
+    image = c.images.detail(pk="...")
+    c.images.download(files=image["files"], filename=Path("path/to/output"))
+
 Credits
 -------
 
--- a/README.rst
+++ b/README.rst
@@ -168,7 +168,7 @@ Downloading Files
 -----------------
 
 An image can consist of one or multiple files, such as a single mha file or a dzi and tiff file. You can download all files
-associated with an image.
+associated with an image at once.
 
 .. code:: python
 
@@ -176,12 +176,16 @@ associated with an image.
 
     downloaded_files = c.images.download(pk="...", filename=Path("path/to/output"))
 
-You can also use other parameters to identify the image, such as the API URL, and you can also supply the "files" list directly.
+You can also use other parameters to identify the image, such as the API URL (use `url="..."`), and you can also supply the "files" list
+directly if you have already obtained the image details.
 
 ... code:: python
 
     image = c.images.detail(pk="...")
     c.images.download(files=image["files"], filename=Path("path/to/output"))
+
+Note that the filename needs to be specified without file extension. The extension is automatically added because multiple files with
+different file extensions can be assosicated with an image (dzi/tif and mhd/zraw for example).
 
 Credits
 -------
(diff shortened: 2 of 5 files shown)


pkcakeout:
You do not access `self.client` directly and yield the calls correctly to the parent that then routes those to the client. Good job! Looks correct to me! And you even added the asynchronous test for testing that the interface works for the async-client, too!

This gets a double-well-done from me I think. :+1: :+1:

> I've implemented this, but I'm struggling a bit to understand the new structure with all these yield calls, so I'm not sure whether I've done that correctly (it works though, at least with the regular client).

I cannot blame you. This is an "invention" from us that differs from other standard solutions that attempt to support both, an async and sync interface for the same library. Typically, this is solved with threads (in "more standard" solutions), which is not an option in our case because it crashes gcapi in some circumstances. And using threads is also considered an anti-pattern... I can go _very deep_ on the explanation here.

Seems like you figured everything out correctly though! If you want to read up on how the async/sync stuff works, open the async_sync_hybrid_support module and read the docstring. I tried to explain why and how this is achieved in gcapi now.

> This method could also be added to the Client class instead of the ImagesAPI, similar to the upload_cases() method.

I actually like that this exists on the ImageAPI. Personally I would probably prefer the `upload_cases()` function to also belong to specific APIs instead where it exists now, but "der Zug ist abgefahren" and we want to preserve backwards compatibility and not randomly break our user-code. So this is good, I think!

jmsmkn:
> @jmsmkn , I think one of the tests randomly failed - seems like some sort of timeout or so maybe? It seems like the docker-pull fails. Any idea why this would be the case?

Yes, you should be able to re-run them, GitHub actions hit ECR anonymous pull limits often.

>  I would probably prefer the upload_cases() function to also belong to specific APIs

It's there as this method cuts across two independent endpoints (cases, uploads), you couldn't add it to cases (say) without knowing about uploads, and that information is only joined on the Client level.

nlessmann:
@pkcakeout thanks a lot! So if I simply interpret `yield` as `await` it is actually relatively straightforward, didn't see that at first. 

Tests are still failing, but I think not the ones that test this feature.

pkcakeout:
>  So if I simply interpret `yield` as `await` it is actually relatively straightforward

Yes, sort-of exactly that. "Sort-of" because we obviously do sth other than "await" with the yield, but that is indeed the intent. And just as a funfact: "await" actually uses the "yield" mechanism itself, and it would also have been possible to write code that drives a standard "async" function. However, the code required for this would be very, very long and convoluted as it pretty much would need to reimplement a big portion of asyncio.

> Tests are still failing, but I think not the ones that test this feature.

Yes, but I get the feeling that the test sequencing might have broken this now. Other tests are also uploading files and maybe there is some sort of assumption on the indexing used for the tests. Since the integration tests manipulate the state of the grand-challenge instances, it is possible that you introducing a test at the _start_ of the file broke a test further down. An easy test to verify this would be to move the test to the end.

I have no time to do this right now but might sneak that in later today.

nlessmann:
Hm, I think the problem is that the "raw_image_upload_session_files" endpoint has been removed on the grand challenge side. See https://github.com/comic/grand-challenge.org/pull/2154/files#diff-06d7fe5ae50e0c2a9d0421335b527825920c663295a087f9f59009055035c0b3L65

So actually, the `UploadSessionFilesAPI` should probably be removed from gcapi (together with the broken tests). Seems easy, I can prepare a PR if you have no time.

pkcakeout:
> Hm, I think the problem is that the "raw_image_upload_session_files" endpoint has been removed on the grand challenge side. See https://github.com/comic/grand-challenge.org/pull/2154/files#diff-06d7fe5ae50e0c2a9d0421335b527825920c663295a087f9f59009055035c0b3L65
> 
> So actually, the `UploadSessionFilesAPI` should probably be removed from gcapi (together with the broken tests). Seems easy, I can prepare a PR if you have no time.

If you could do that, that would be awesome. I am running super late on the feature that should be delivered for CIRRUS by the end of next week.

nlessmann:
See https://github.com/DIAGNijmegen/rse-gcapi/pull/97

PR #97, opened 2021-11-12
Remove UploadSessionFilesAPI

nlessmann:
Since the corresponding endpoint has been removed (I suppose because files are now directly uploaded to S3) in https://github.com/comic/grand-challenge.org/pull/2154 I believe the UploadSessionFilesAPI can be removed.


jmsmkn (approved):
Thanks! Oversight from me, indeed, this endpoint was deprecated and is now removed.

PR #100, opened 2022-01-19
Add option to specify interface in upload_cases

amickan:
This addresses part 1 of https://github.com/DIAGNijmegen/rse-roadmap/issues/114

Associated GC PRs: 
https://github.com/comic/grand-challenge.org/pull/2249
https://github.com/comic/grand-challenge.org/pull/2259

I think the tests can only pass once the GC PR has been merged, right?

### DIFF
--- a/gcapi/client.py
+++ b/gcapi/client.py
@@ -575,30 +575,36 @@ class ClientBase(ApiDefinitions, ClientInterface):
             )
         )
 
-    def upload_cases(
+    def upload_cases(  # noqa: C901
         self,
         *,
         files: List[str],
         archive: str = None,
         reader_study: str = None,
         answer: str = None,
+        archive_item: str = None,
         interface: str = None,
     ):
         """
-        Uploads a set of files to an archive or reader study.
+        Uploads a set of files to an archive or reader study, or a single
+        file to an archive item.
         A new upload session will be created on grand challenge to import and
-        standardise your files. This function will return this new upload
+        standardise your file(s). This function will return this new upload
         session object, that you can query for the import status. If this
-        import is successful, the new images will then be added to the selected
-        archive or reader study.
-        You will need to provide the slugs of the objects to pass the images
-        along to. You can find this in the url of the object that you want
+        import is successful, the new image(s) will then be added to the selected
+        archive, archive item or reader study.
+        You will need to provide the slugs of the archive or reader study or the
+        pk of the archive item to pass the images along to.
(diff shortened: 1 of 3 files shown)


gcapi/client.py
```diff
@@ -575,30 +575,36 @@ def _upload_files(self, *, files, **kwargs):
             )
         )
 
-    def upload_cases(
+    def upload_cases(  # noqa: C901
```
jmsmkn: Is it possible to simplify this rather than ignoring the too-complex error?

gcapi/client.py
```diff
@@ -634,19 +643,33 @@ def upload_cases(
...
             raise ValueError(
-                "An interface can only be defined for archive uploads."
+                "An interface can only be defined for archive and archive item "
+                "uploads."
             )
 
+        if archive_item:
+            if len(files) > 1:
+                raise ValueError(
+                    "You can only upload one file to an archive item at a time."
```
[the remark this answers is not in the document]
amickan: No, this is indeed not true, but it's also no longer in the code. This was in an older version of the PR...

gcapi/client.py
amickan: and this is also no longer in there... strange.

gcapi/client.py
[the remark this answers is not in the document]
jmsmkn: Oh, this is really strange! GitHub does indeed say that 2 new commits were added after I last viewed it, _then_ shows my review after those. I wonder what went wrong there, I thought I double-checked the view all change option, but maybe not, has been some long days. The best kind of review comments are ones that are already solved, well done :-)

gcapi/client.py
```diff
@@ -577,20 +582,31 @@ def upload_cases(
...
         the slug for this is "corads-ai", so you would call this function with
             upload_cases(files=[...], archive="corads-ai")
+        For archive uploads, you can additionally specify the interface slug
+        for the to be created archive items. For archive item uploads, the
+        interface is mandatory. If you define an already existing interface,
+        the old file associated with that interface will be replaced by the new file.
+        You can find a list of interfaces here:
+        https://grand-challenge.org/algorithms/interfaces/
+        The interface slug corresponds to the lowercase hyphenated title of the
+        interface, e.g. generic-medical-image for Generic Medical Image.
```
jmsmkn, line 609: We should add the slug to the table to help users as this is not always the case.

jmsmkn (approved):
LGTM!

PR #102, opened 2022-02-09
Enable archive item editing

amickan:
Enables users to add and update (non-)image files and values to an existing archive item. 
Specifying an interface slug that already exists for the archive item, will overwrite the existing value. Specifying a new interface will add a civ with that interface and the given value. 
For existing image civs, users can also change the interface kind. 

Related GC PR: https://github.com/comic/grand-challenge.org/pull/2268


gcapi/client.py
```diff
@@ -741,3 +747,100 @@ def run_external_job(self, *, algorithm: str, inputs: Dict[str, Any]):
...
+        Returns
+        -------
+        The updated archive item
+        """
+        item = yield from self.__org_api_meta.archive_items.detail(
+            pk=archive_item_pk
+        )
+        civs: Dict[str, list] = {"values": []}
+
+        for civ_title, value in values.items():
```
jmsmkn, line 801: Aren't these are `ci_slugs`, not `civ_titles`? We should never use titles as they're not unique.

gcapi/client.py
```diff
@@ -232,10 +233,14 @@ class ArchivesAPI(APIBase):
     base_path = "archives/"
 
 
-class ArchiveItemsAPI(APIBase):
+class ArchiveItemsAPI(ModifiableMixin, APIBase):
     base_path = "archives/items/"
 
 
+class ComponentInterfacesAPI(APIBase):
+    base_path = "interfaces/"
```
jmsmkn, line 241: Remember to update the base if we change it.

gcapi/client.py
```diff
@@ -741,3 +747,100 @@ def run_external_job(self, *, algorithm: str, inputs: Dict[str, Any]):
...
+                raise ValueError(
+                    f"{civ_title} is not an existing interface. "
+                    f"Please provide one from this list: "
+                    f"https://grand-challenge.org/algorithms/interfaces/"
+                ) from e
+            i = {"interface": ci["slug"]}
+
+            if not value:
+                raise ValueError(
+                    f"You need to provide a value for {ci['title']}"
```
jmsmkn, line 816: Use `slug` here.

jmsmkn (commented):
Looks great!

jmsmkn (approved):

PR #104, opened 2022-02-14
Update pre commit versions

jmsmkn:


gcapi/exceptions.py
```diff
@@ -2,9 +2,9 @@ class GCAPIError(Exception):
     """Base class for all exceptions"""
 
 
-class ObjectNotFound(GCAPIError):
+class ObjectNotFound(GCAPIError):  # noqa: N818
```
chrisvanrun, line 5: Likely but curious: is not adding the suffix `Error` here done for backward compatibility?

chrisvanrun (approved):
LGTM

PR #108, opened 2022-02-22
CI Standardisation

jmsmkn:
This PR:
- Migrates to PEP517
- Standardises README and HISTORY formats to markdown
- Removes codecov (didn't work, we can discuss how to handle that service)


chrisvanrun:
Seem like the tox env is mixing up the python version calls (i.e. py38 is being called in py37)

.github/workflows/ci.yml
```diff
@@ -1,48 +1,73 @@
...
-    needs: [precommit]
+  build:
     runs-on: ubuntu-latest
     strategy:
       matrix:
         python: ["3.6", "3.7", "3.8", "3.9", "3.10"]
     steps:
       - uses: actions/checkout@v2
-      - name: Setup Python
+      - name: Set up Python ${{ matrix.python-version }}
```
chrisvanrun, line 41: Possibly this should be `matrix.python`?

jmsmkn: Good catch! The usage seems OK, but this also happens on other repos where we test for 3.10 and use strings for the python version.

jmsmkn: You were right! Should be resolved now.

chrisvanrun (approved):
LGTM!

PR #117, opened 2022-03-08
Missed a schema (again)

miriam-groeneveld:


jmsmkn (approved):
Maybe good to install this into CIRRUS from github rather than pypi to ensure that nothing else it missed.

PR #122, opened 2022-03-22
add display_sets to ReaderStudyAPI

miriam-groeneveld:


jmsmkn:
I don't know what is going on with the ECR public connectivity in CI here.

miriam-groeneveld:
No idea either but can it be this issue: https://github.com/aws/aws-cli/issues/4962#issuecomment-586486939 
Not sure if it has anything to do with moving gc to Ireland?

jmsmkn:
Yes the region is definitely set correctly - this wasn't affected by the move to eu-west-1 as ECR public is only in us-east-1. The grand challenge push is flaky, is sometimes does succeed (e.g. 1 hour ago https://github.com/comic/grand-challenge.org/runs/5660381960?check_suite_focus=true). My guess is that the CI for this repo just hits it more often. I have turned off fail fast to see if we can get some things passing.

jmsmkn:
Chris Russ and I looked into this a bit. I cannot reproduce the TLS Handshake timeouts on any other machine, so I think this is a problem with GHA rather than ECR. GitHub also announced this incident so my gut feeling is that it is all connected https://www.githubstatus.com/incidents/tyc8wpsgr2r8. We have the fail fast so can manually retry things for now, but may need to add an auto retry.

PR #126, opened 2022-03-28
use pk instead of id for archives

miriam-groeneveld:
Fixes failing tests due to a change in the way archives are serialized (with pk instead of id)


jmsmkn:
I added this in e8af98add1633ddf2610fae2723baabacee6787b and also the change in the image key in 6488de2edc487e54643816e7e6bdb70ac09801f6.

PR #127, opened 2022-04-07
Fix not being able to upload files using string file paths

chrisvanrun:
Ran into this when testing out a helper function.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -62,7 +62,7 @@ jobs:
           retry() {
             local -r -i max_attempts="$1"; shift
             local -i attempt_num=1
-            until ("$@")
+            until eval "$@"
             do
               command_return=$?
               if ((attempt_num==max_attempts))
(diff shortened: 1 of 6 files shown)


gcapi/client.py
```diff
@@ -578,7 +578,7 @@ def _upload_files(self, *, files, **kwargs):
                 uploads.append(
                     (
                         yield from self.__org_api_meta.uploads.upload_fileobj(
-                            fileobj=f, filename=file.name
+                            fileobj=f, filename=f.name
```
chrisvanrun, line 581: This is the actual fix, the rest is tests to cover both `str` and `Path` file uploads.

jmsmkn:
This looks good but extending those tests is really slow, could you instead write a dedicated test for this?

chrisvanrun:
Changes:
* Added some lines about testing in README
* Added a simple bash-level retry for pulling images (seems like the login was the main culprit here)
* Simplified testing for this one change

jmsmkn:
Pinging @pkcakeout to review, especially the retry stuff

.github/workflows/ci.yml
```diff
@@ -58,12 +58,27 @@ jobs:
...
-          aws ecr-public get-login-password --region us-east-1 | docker login --username AWS --password-stdin public.ecr.aws/diag-nijmegen/grand-challenge/http
-          docker pull public.ecr.aws/diag-nijmegen/grand-challenge/http:latest
-          docker logout public.ecr.aws/diag-nijmegen/grand-challenge/http
+          retry() {
+            local -r -i max_attempts="$1"; shift
+            local -i attempt_num=1
+            until eval "$@"
+            do
+              command_return=$?
+              if ((attempt_num==max_attempts))
```
pkcakeout, line 67: Would do >= instead of == because that better captures intent.

.github/workflows/ci.yml
```diff
@@ -58,12 +58,27 @@ jobs:
...
+              then
+                return $command_return
+              else
+                echo "Retrying..."
+                sleep $((attempt_num++ * 10))
+              fi
+            done
+          }
```
pkcakeout, line 75: I think this is probably correct, but I wonder why we are not using one of the solutions that are available right there - either on the commandline or on the marketplace for GitHub Actions? I found a few from a quick googling in the marketplace - the first one would probably already work:

- https://github.com/marketplace/actions/shell-command-retry-action
- https://github.com/marketplace/actions/retry-action#retry-command
- https://github.com/marketplace/actions/retry-step

My favorite solution however, would be to just use `retry` om thje console:

~~~~~~~~~~~~
root@2e7d0e40a0c9:/# apt install retry
[...]
root@2e7d0e40a0c9:/# retry --help


NAME
  retry - Repeat command until a criteria is met, usually success.

SYNOPSIS
  retry [-v] [-h] [-u until] [-w while] command ...

DESCRIPTION

  The tool repeats the given command until the command is successful,
  backing off with a configurable delay between each attempt.

[...]
~~~~~~~~~~~~

pkcakeout:
I can fix this next week...

.github/workflows/ci.yml
pkcakeout: PR #130 contains my fix for this...

jmsmkn (approved):

PR #128, opened 2022-04-19
Provide access to the page meta information

chrisvanrun:
When calling the `.page()` of the APIBase I have a need to access the returned `"count"`.

I could get away with doing a custom page call via URL components to get these values, but then spoofing them in CIRRUS becomes a bit of a hassle and a mess.

The quickest way I could come up with, without changing the return values was to park the page meta data into a referenced dictionary.

### DIFF
--- a/gcapi/apibase.py
+++ b/gcapi/apibase.py
@@ -88,27 +88,33 @@ class APIBase(Common):
             self.verify_against_schema(i)
         return result
 
-    def page(self, offset=0, limit=100, params=None):
+    def page(self, offset=0, limit=100, params=None, page_meta=None):
         if params is None:
             params = {}
         params["offset"] = offset
         params["limit"] = limit
-        result = (
-            yield self.yield_request(
-                method="GET", path=self.base_path, params=params
-            )
-        )["results"]
-        for i in result:
+        response = yield self.yield_request(
+            method="GET", path=self.base_path, params=params
+        )
+        results = response["results"]
+        for i in results:
             self.verify_against_schema(i)
-        return result
+        if page_meta is not None:
+            page_meta.update(
+                {k: v for k, v in response.items() if k != "results"}
+            )
+        return results
 
     @mark_generator
-    def iterate_all(self, params=None):
+    def iterate_all(self, params=None, page_meta=None):
         req_count = 100
         offset = 0
         while True:
             current_list = yield from self.page(
-                offset=offset, limit=req_count, params=params
+                offset=offset,
+                limit=req_count,
+                params=params,
+                page_meta=page_meta,
             )
             if len(current_list) == 0:
                 break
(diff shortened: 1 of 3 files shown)


chrisvanrun:
CI is failing because of infrastructure issues. These are addressed in:
- #127

pkcakeout (commented):
Oof... I think passing a dict sort-of "by-reference" works, of course, but is a bit of an anti-pattern.

But I would have a proposal for a different implementation, I think is a lot cleaner: Attach the metadata to _everything by default:

~~~~~~~
class PageResult(dict):
    offset: int = 0
    count: int = 0
    # ??? page: int
    
    def __init__(self, offset, count, *args, **kwargs):
        super(PageResult, self).__init__(*args, **kwargs)
        self.offset = offset
        self.count = count

#---------

Then later:

response = yield self.yield_request(
    method="GET", path=self.base_path, params=params
)
results = PageResult(response["offset"], response["count"], response["results"])
~~~~~~~

This way we avoid having to use mutable arguments which are super-messy. We do not change the interface and make the relevant data accessible everywhere. Just provide a positive list of data elements you want to extract. Right now you extract everything except "results" which could also be considered messy because there is no positive data validation happening. We validate everything else through the use of json schema. Would be nice to have the pagination metadata also be validated by providing the list of elements you are actually interested in.

chrisvanrun:
> This way we avoid having to use mutable arguments which are super-messy.
Quick haste generally makes a mess.

I've incorporated your suggestion. I didn't see any neat way to also cover `iterate_all` and ended-up dropping that. Rational is if one decides to just iterate over all items instead of paginating yourself I suspect you don't really care about the total count and you can bet that anyway by retrieving a `.page(limit=1)`.

pkcakeout:
Regarding iterate_all: Yes pretty much what you say. You _could_ create a conditional wrapper to also pass the data along, but I do not expect that this will be every used for the reason you are stating yourself :-)

gcapi/apibase.py
```diff
@@ -176,3 +182,15 @@ def partial_update(self, pk, **kwargs):
 
     def delete(self, pk):
         return (yield from self.perform_request("DELETE", pk=pk))
+
+
+class PageResult(list):
```
pkcakeout, line 187: Just out of curiosity. Did you find any references on subclassing from list vs UserList?

pkcakeout (approved):

gcapi/apibase.py
chrisvanrun: It was a bit of a mixed bag. Python2 had a strict suggestion to use UserList, this is not the case for python 3. 

Depending on what behaviour you are looking for the common denotion is now to use a collections (e.g. UserList, or Sequence). I felt it should be a minimum change of the interface so I went with `list`. This did mean that the `count()` couldn't be overwritten, hence the `total_count`.

PR #132, opened 2022-06-13
Update conftest for recent Grand Challenge development changes

jmsmkn:
`docker compose` doesn't work on GitHub actions right now, see comments.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -45,15 +45,28 @@ def local_grand_challenge() -> Generator[str, None, None]:
                 "scripts/development_fixtures.py",
                 "scripts/algorithm_evaluation_fixtures.py",
                 "scripts/image10x10x10.mha",
+                "scripts/minio.py",
                 "app/tests/resources/gc_demo_algorithm/copy_io.py",
                 "app/tests/resources/gc_demo_algorithm/Dockerfile",
             ]:
                 get_grand_challenge_file(Path(f), Path(tmp_path))
 
             try:
-                check_call(["make", "development_fixtures"], cwd=tmp_path)
                 check_call(
-                    ["make", "algorithm_evaluation_fixtures"], cwd=tmp_path
+                    [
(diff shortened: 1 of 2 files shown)


tests/conftest.py
```diff
@@ -45,15 +45,28 @@ def local_grand_challenge() -> Generator[str, None, None]:
                 "scripts/development_fixtures.py",
                 "scripts/algorithm_evaluation_fixtures.py",
                 "scripts/image10x10x10.mha",
+                "scripts/minio.py",
```
chrisvanrun, line 48: I assume an unrelated fix for the problem at hand?

tests/conftest.py
```diff
@@ -122,4 +138,22 @@ def rewrite_docker_compose(content: bytes) -> bytes:
...
     return yaml.safe_dump(spec).encode("utf-8")
+
+
+def rewrite_makefile(content: bytes) -> bytes:
+    # Using `docker compose` with version 2.4.1+azure-1 does not seem to work
+    # It works locally with version `2.5.1`, so for now go back to docker-compose
+    # If this is fixed docker-compose-wait can be removed and the `--wait`
+    # option added to the "up" action above
+    makefile = content.decode("utf-8")
+    makefile = makefile.replace("docker compose", "docker-compose")
```
chrisvanrun, line 158: This will work for now, but care should be taken for in some configurations/OS/frameworks `docker-compose` is just an alias for `docker compose`.

chrisvanrun (approved):
LGTM. Small note of care on aliasing.

tests/conftest.py
jmsmkn: This is now required for `make development_fixtures`, was changed a couple of weeks ago.

tests/conftest.py
jmsmkn: Yep, we should hear about it if it breaks.

PR #134, opened 2022-07-11
Remove POST Validation

jmsmkn:
Grand Challenge will do this validation for the client, no need
to repeat it.


pkcakeout (approved):
Makes sense :+1:

PR #135, opened 2022-07-13
Add ANGLE and MULTIPLE_ANGLE answer schemas

HarmvZ:


jmsmkn:
Could you add this to grand challenge first along with the tests to ensure that the types are unique?

PR #137, opened 2022-08-19
Allow algorithm jobs for file super type interfaces

MikeOverkamp-diag:


amickan:
I think the algorithm fixtures on GC still need changes for the tests you added to work.

MikeOverkamp-diag:
You're right, I forgot to commit them!

amickan (approved):

PR #140, opened 2022-09-08
Remove json schemas

chrisvanrun:
The gcapi answer JSON schema and reader-study schema are out of sync with that of grand-challenge.

For instance, "name" is not a required property for some annotations: https://github.com/comic/grand-challenge.org/blob/1e234be514cb668a65194c080d6411cbc6305550/app/grandchallenge/components/schemas.py#L97

Edit: This PR removes all schemas and related code/depencencies.


jmsmkn:
What does @pkcakeout think? We keep coming back to this. See previous discussions (#134, https://teams.microsoft.com/l/message/19:54cf52330bfb4c33991e550b49355741@thread.tacv2/1655815817296?tenantId=b208fe69-471e-48c4-8d87-025e9b9a157f&groupId=45c973c4-5543-45ad-a8b8-79ea596e6c69&parentMessageId=1655814054739&teamName=Research%20Software%20Engineering&channelName=General&createdTime=1655815817296&allowXTenantAccess=false)

jmsmkn:
This is probably the best thing we can do until we can use an up-to-date version of Pydantic in MeVisLab.

pkcakeout:
I think we can afford to remove them. Afaik, we are still using my cryptic, but robust DSL-based to extract the annotation data:

https://github.com/DIAGNijmegen/rse-cirrus-core/blob/7ee70985bdcc29a9d9d9a2f80ca50b7576b574a0/src/cirrus/DIAG/CIRRUSCore/Modules/Scripts/python/cirrus/workstations/core_web/annotations/cso_serialization.py#L419-L422

which _again_ does verification, so this still will be checked on the data-ingestion side. Not as thoroughly as it used to with schemas, but it checks what is necessary for CIRRUS, which is what we need.

So :+1:

jmsmkn:
Then lets remove all the schemas, the whole mechanism and the dependencies.

pkcakeout:
Oh yeah, did not realize that those are the last schemas that are in there in the first place :D Well, ok. Then you got your way afterall :wink: 

But yes, agreed, let's delete the feature then! We add the safety railing elsewhere - probably `grand_challenge_interface` for CIRRUS.

chrisvanrun:
> Oh yeah, did not realize that those are the last schemas that are in there in the first place :D Well, ok. Then you got your way afterall 😉

Not fully, IIRC there are some stale (but still in use schemas) which are uniquely checked via gcapi. I'll look into moving those to GC.

jmsmkn (approved):

PR #142, opened 2022-09-29
Add automatic retries for requests

chrisvanrun:
Related:
- https://github.com/DIAGNijmegen/rse-cirrus-core/issues/1675

This PR adds retries for requests made to grand challenge.

~A client will retry `403` (_Forbidden_), and `404` (_Not Found_) once without delay. The rationale is that there might be a split-second delay in permission propagation that is covered by regular latency.~ Edit: it now only retries 5xx.

Several transient server errors (i.e. `5xx`) are retried with an exponential backoff delay in seconds: 0.1, 0.2, 0.4, 0.8, 1.60, 3.20, 6.40, 12.8. With a typical total retry time of ~26s., but the exact retry duration depends on the type of responses and their order.

A **custom** retry strategy can be provided to the client as long as a call of the provided strategy results in an object that inherits from the `BaseRetryStrategy` and implements a `get_delay(response)` method.

### DIFF
--- a/gcapi/retries.py
+++ b/gcapi/retries.py
@@ -1,0 +1,7 @@
+import abc
+
+
+class RetryStrategy(abc.ABC):
+    def get_interval_ms(self, response):
+        """Returns the number of ms to pause before the next retry given the latest response"""
+        pass
--- a/gcapi/retries.py
+++ b/gcapi/retries.py
@@ -1,11 +1,10 @@
-import abc
-
-
-class RetryStrategy(abc.ABC):
-    def get_interval(self, response):
+class BaseRetries:
+    def get_delay(self, latest_response):
         """
         Returns the number of seconds to pause before the next retry, based on the latest response.
 
         Optionally, if None is returned no retry is to be performed.
         """
-        pass
+        raise NotImplementedError(
+            "The 'get_delay' method must be implemented."
+        )
--- a/gcapi/retries.py
+++ b/gcapi/retries.py
@@ -1,10 +1,71 @@
-class BaseRetries:
+from httpx import codes
+
+
+class BaseRetryStrategy:
     def get_delay(self, latest_response):
         """
-        Returns the number of seconds to pause before the next retry, based on the latest response.
+        Returns the number of seconds to pause before the next retry,
+        based on the latest response.
 
         Optionally, if None is returned no retry is to be performed.
         """
         raise NotImplementedError(
             "The 'get_delay' method must be implemented."
+        )  # pragma: no cover
+
+
+NO_RETRY = None
+RETRY_NOW = 0
+
+
+class SelectiveBackoffStrategy(BaseRetryStrategy):
+    """
+    Retries responses with 403 and 404 once and several transient
+    server errors (i.e. 5xx) with an exponential backoff.
+
+    Each response code has its own backoff counter
+
+    """
+
+    def __init__(self, backoff_factor, maximum_number_of_retries):
+        self.backoff_factor = backoff_factor
+        self.maximum_number_of_retries = maximum_number_of_retries
+        self.earlier_number_of_retries = dict()
+
+    def __call__(self):
+        return self.__class__(
+            backoff_factor=self.backoff_factor,
+            maximum_number_of_retries=self.maximum_number_of_retries,
         )
+
+    def get_delay(self, latest_response):
+        for handling_codes, strategy in self.strategies.items():
+            if latest_response.status_code in handling_codes:
+                return strategy(self, latest_response.status_code)
+        return NO_RETRY
+
+    def _single_retry(self, code: int):
+        if code in self.earlier_number_of_retries:
+            return NO_RETRY
+        else:
+            self.earlier_number_of_retries[code] = 1
+            return RETRY_NOW
+
+    def _backoff_retries(self, code: int):
+        num_retries = self.earlier_number_of_retries.get(code, 0)
+        if num_retries >= self.maximum_number_of_retries:
+            return NO_RETRY
+        else:
+            self.earlier_number_of_retries[code] = num_retries + 1
+            return self.backoff_factor * (2**num_retries)
+
+    strategies = {
+        (codes.FORBIDDEN, codes.NOT_FOUND): _single_retry,
+        (
+            codes.INTERNAL_SERVER_ERROR,
+            codes.BAD_GATEWAY,
+            codes.SERVICE_UNAVAILABLE,
+            codes.GATEWAY_TIMEOUT,
+            codes.INSUFFICIENT_STORAGE,
+        ): _backoff_retries,
+    }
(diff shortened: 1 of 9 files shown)


jmsmkn:
> A client will retry 403 (Forbidden), and 404 (Not Found) once without delay. The rationale is that there might be a split-second delay in permission propagation that is covered by regular latency.

That definitely should not be done, permissions are updated in the same transaction.

chrisvanrun:
> That definitely should not be done, permissions are updated in the same transaction as objects being created, they're not deferred.

Good to know: updated!

chrisvanrun:
> > A client will retry 403 (Forbidden), and 404 (Not Found) once without delay. The rationale is that there might be a split-second delay in permission propagation that is covered by regular latency.
> 
> That definitely should not be done, permissions are updated in the same transaction as objects being created, they're not deferred.

Just to check: this also holds for adding ppl to groups (i.e. adding a reader to a reader study?).

jmsmkn:
> Just to check: this also holds for adding ppl to groups (_e.g._ adding a reader to a reader study)?

Yes. Users are never assigned permissions directly, all done through groups, and all done in the same transaction.

tests/utils.py
```diff
@@ -31,3 +32,50 @@ async def wrapper(*args, **kwargs):
...
+                ResponseMetaData.num_requests += 1
+                response = responses.__next__()
+                response.request = request
+                return response
+            except StopIteration as e:
+                raise RuntimeError("Ran out of mock responses") from e
+
+        async def handle_async_request(self, *args, **kwargs):
+            return self.handle_request(*args, **kwargs)
+
```
pkcakeout, line 65: Did you find this doc-entry: https://www.python-httpx.org/advanced/#mock-transports ?

It points to a built-in solution which probably is more robust down the line, but also points to fully-fledged mocking solutions for, for example, pytest, which I guess would be even handier. Typically, you would need to rewrite your test to query different virtual endpoints, but I think that would be worthwhile for a more robust solution?!

pkcakeout (commented):
I do bit know enough about HTTPTransports to really understand how this works wrt object lifetimes etc. I have to trust that it works as expected... If someone wants me to make a real in-depth review with digging into httpx and verifying that this matches the library, please tell me. Other than that, I understand the intent, and I can _imagine_ how it works and it pretty much makes sense to me!

----------------

Other than that, and more importantly: I think it makes sense from a user-perspective to have this. But I will let @jmsmkn greenlight this because I see the possibility of this making "emergencies" a lot more severe. Even with exponential backoff, this will increase the number of (failing) requests to grand-challenge dramatically. Students, we, everyone uses gcapi to interface with grand-challenge.

I post this a comment only. There are some suggested fixes though!

gcapi/transports.py
```diff
@@ -0,0 +1,121 @@
...
+    governed by a retry strategy.
+
+    Once a request fails, a retries-strategy object is created from the provided
+    retries class. This object is then queried to provide the delay in seconds
+    via 'get_delay(response)'.
+
+    The retry object should be an instance of BaseRetries.
+    """
+
+    @wraps(httpx.HTTPTransport.handle_request)
```
pkcakeout, line 66: Forgot to mark this. This is also not a `@wraps` case.

gcapi/transports.py
```diff
@@ -0,0 +1,121 @@
...
+                response.close()
+                sleep(retry_delay)
+
+        return response
+
+
+class AsyncRetryTransport(BaseRetryTransport, httpx.AsyncHTTPTransport):
+    """Same as the RetryTransport but adapted for asynchronous clients"""
+
+    @wraps(httpx.AsyncHTTPTransport.handle_async_request)
```
pkcakeout, line 96: No `@wraps`

gcapi/retries.py
```diff
@@ -0,0 +1,64 @@
...
+        self.maximum_number_of_retries: int = maximum_number_of_retries
+        self.earlier_number_of_retries: Dict[int, int] = dict()
+
+    def __call__(self) -> BaseRetryStrategy:
+        return self.__class__(
+            backoff_factor=self.backoff_factor,
+            maximum_number_of_retries=self.maximum_number_of_retries,
+        )
+
+    @wraps(BaseRetryStrategy.get_delay)
```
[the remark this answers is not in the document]
chrisvanrun: Since Python doesn't really do interfacing, the `@wrap` was just meant to copy over the docstring and annotations. Having played around a bit, it seems like the doc strings are actually pass through. Or at least `help()` returns the overwritten one.

gcapi/transports.py
```diff
@@ -0,0 +1,121 @@
...
+    ):
+        super().__init__(*args, **kwargs)
+        self.retry_strategy = retry_strategy
+        if self.retry_strategy:
+            obj = self.retry_strategy()
+            if not isinstance(obj, BaseRetryStrategy):
+                raise RuntimeError(
+                    "Provided retries strategy should be None "
+                    "or when called produce an instance of BaseRetry"
+                )
```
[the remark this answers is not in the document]
chrisvanrun: I prefer the simpler approach of instantiating it once.

chrisvanrun:
> I do bit know enough about HTTPTransports to really understand how this works wrt object lifetimes etc. I have to trust that it works as expected... If someone wants me to make a real in-depth review with digging into httpx and verifying that this matches the library, please tell me. Other than that, I understand the intent, and I can _imagine_ how it works and it pretty much makes sense to me!
> 
> Other than that, and more importantly: I think it makes sense from a user-perspective to have this. But I will let @jmsmkn greenlight this because I see the possibility of this making "emergencies" a lot more severe. Even with exponential backoff, this will increase the number of (failing) requests to grand-challenge dramatically. Students, we, everyone uses gcapi to interface with grand-challenge.
> 
> I post this a comment only. There are some suggested fixes though!

I've tried to flow the docs that describe roughy how the transports are to look and work.

gcapi/retries.py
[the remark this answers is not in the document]
pkcakeout: > Or at least help() returns the overwritten one.

Yes exactly! That was my point. They are not really "pass-through", it is a bit more complicated... the class-system uses a separate mechanism, which you actually you already reminded me of, with `mro`-stuff. :+1:

tests/utils.py
pkcakeout: No probs... if there is a reason to do this, it is all fine!

jmsmkn:
> Other than that, and more importantly: I think it makes sense from a user-perspective to have this. But I will let @jmsmkn greenlight this because I see the possibility of this making "emergencies" a lot more severe. Even with exponential backoff, this will increase the number of (failing) requests to grand-challenge dramatically. Students, we, everyone uses gcapi to interface with grand-challenge.

So long as it works it's fine, I've seen many people just do `for thin in things: try: ... except: pass` with gcapi.

pkcakeout (approved):
I already did a review, and since james is fine with it, here we go!

PR #147, opened 2023-02-17
Update Python Support

jmsmkn:
Removes support for Python 3.6 and 3.7, adds support for Python 3.11.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,7 +8,7 @@ on:
       - published
 
 env:
-  MINIMUM_PYTHON_VERSION: '3.6'
+  MINIMUM_PYTHON_VERSION: '3.7'
 
 concurrency:
   group: ${{ github.head_ref || github.run_id }}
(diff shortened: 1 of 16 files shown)


jmsmkn:
@pkcakeout Do you know how this would change now that we can finally use dataclasses? https://github.com/DIAGNijmegen/rse-gcapi/blob/3fa0bcb4c1c16e80567d64ae42cef9cf585f93ff/gcapi/sync_async_hybrid_support.py#L171-L174

pkcakeout:
> @pkcakeout Do you know how this would change now that we can finally use dataclasses?
> 
> https://github.com/DIAGNijmegen/rse-gcapi/blob/3fa0bcb4c1c16e80567d64ae42cef9cf585f93ff/gcapi/sync_async_hybrid_support.py#L171-L174
> 
> This was my guess [5466285](https://github.com/DIAGNijmegen/rse-gcapi/commit/546628549f15df893f101176f5e233bdc1f81694)

I think that is it! I actually had a version that was using dataclasses, but I think it is lost. But like I said: I think this was just it!

pkcakeout (approved):
Lot's of noise because of the black update, I guess?! Anyway. I did not see anything crazy or wrong...

PR #151, opened 2023-06-20
Do not retry 500 errors

jmsmkn:
500 errors mean that the request the client is making is definitely bad, and is results in a backend error on Sentry. Right now, if users create such a request it will hammer Grand Challenge and Sentry with errors when the problem is with the clients request.


chrisvanrun (approved):
Makes sense.

jmsmkn:
OK maybe not hammering but it was still 500 events / hour for 2.5 hours from a single user.

PR #154, opened 2024-01-23
Remove retina endpoints

jmsmkn:
See https://github.com/DIAGNijmegen/rse-grand-challenge-admin/issues/236


jmsmkn:
The tests on main are likely broken.

tests/async_integration_tests.py
```diff
@@ -170,20 +73,22 @@ async def test_chunked_uploads(local_grand_challenge):
...
-    async with AsyncClient(
-        token=RETINA_TOKEN, base_url=local_grand_challenge, verify=False
-    ) as c_retina:
-        existing_chunks_retina = (await c_retina(path="uploads/"))["count"]
+    # archive
+    async with (
+        AsyncClient(
+            token=ARCHIVE_TOKEN, base_url=local_grand_challenge, verify=False
+        ) as c_archive
+    ):
```
HarmvZ, line 81:
```suggestion
    async with AsyncClient(
        token=ARCHIVE_TOKEN, base_url=local_grand_challenge, verify=False
    ) as c_archive:
```

HarmvZ (changes requested):

HarmvZ:
@jmsmkn I just came across this. Do you know what is going on with the failing tests?

jmsmkn:
Likely something with the GC fixtures changed and the tests broke. They should be re-run on main. If so, it is not ideal to rely on those fixtures kept on GC. Maybe we should create the fixtures that are needed for the GCAPI tests within this repo.

HarmvZ (approved):
Right, I just ran the tests on main and they are also broken. Issue here: https://github.com/DIAGNijmegen/rse-gcapi/issues/155. Let's merge this changeset since the failing tests are unrelated.

jmsmkn:
I'm not sure if this PR works, I think it's better to fix the tests on main first then merge/rebase.

HarmvZ:
Sidenote: Some references to retina stuff are still in the models.py file. This will automatically be removed when we regenerate the file.

PR #156, opened 2024-07-26
Fix broken tests

HarmvZ:
**Warning: This PR is based on the remove_retina branch**

I started by seeing when the tests did pass and if that is still the case currently. Looking at the latest merge it was around june 20 2023 so I started with gc around there (`6ef429ae7691afcadfbfae6e1da834ce0e3ba76b`), this passes. Then I iterated over some releases to see when the first errors pop up. This happened at `cc514a444386d1787a7b4ef5d1395c22e33df07d`.

Reason: Faker is added [here](https://github.com/comic/grand-challenge.org/pull/3093) but not added in poetry. It is available in the test container because it is a dependency of factory-boy, however it is not available in the production container, which is the container used by gcapi to run a local gc in the tests. This needs to be fixed by either adding it in the production container or removing the usage in the fixtures. For now, the easy fix was to "hack it in":

An algorithm with file as input was used in the tests but it is removed here: https://github.com/comic/grand-challenge.org/pull/3140. I've added it back in a local version of algorithm_evaluation_fixtures.py.
Also in that PR, the demo algorithm used in the fixtures is built using docker. Docker is not available in the production container, so in the local version of evaluation fixtures we load the algorithm from the local filesystem.

Functionality for updating CIV interface by uploading existing file is removed here: https://github.com/comic/grand-challenge.org/pull/3156. I've removed the tests and some documentation for this functionality.

At this point the codebase was at the removal of the retina endpoints: https://github.com/comic/grand-challenge.org/pull/3192. This obviously introduced errors because of the missing endpoints. After merging #154, those errors dissapear, except for 2 new errors. Those are fixed by switching a token and removing string interpolation issue.

Then, after updating to the most recent version of gc, there were 2 issues remaining: 
- 1 caused by a deprecated property on the Job model. This could also be fixed by regenerating the models.py file with datamodel-code-generator, but this was an easy fix so I just removed the property.
- component_interface_value_fixtures.py was missing, added it back.

After all this the tests pass again 🎉 

Some of the fixes here are hacky, we need to find a proper solution for them. Also we need to find a proper solution for keeping gcapi up to date with gc because otherwise changes will creep in that we miss here. Then we have broken version of gcapi and run into failing tests when we create a PR here. Maybe the tests here should be run periodically (weekly/bimonthly?) so that we find new errors sooner.


HarmvZ:
It seems there is one flaky test remaining. See this test run: https://github.com/DIAGNijmegen/rse-gcapi/actions/runs/10113674491/attempts/1

I reran the workflow without changing anything and then it passed.

jmsmkn:
I think it would be best if we completely decouple the two repositories. These problems pop up often as the Grand Challenge codebase changes a lot, and we do not have automated tests for drifts. I know the tests well so am happy to do that today.

gcapi/client.py
```diff
@@ -723,20 +723,6 @@ def update_archive_item(
...
-        also change the interface type (e.g. from generic medical image to
-        generic overlay) by providing the link to the existing image together
-        with the new interface slug you would like to use:
-        client.update_archive_item(
-            archive_item_pk=items[0]['id'],
-            values={
-                "generic-overlay":
-                "https://grand-challenge.org/api/v1/cases/images/.../",
-            }
-        )
```
jmsmkn, line 739: Why was this help text removed?

HarmvZ: This functionality has been removed in gc, see https://github.com/comic/grand-challenge.org/pull/3156.

jmsmkn: Shouldn't that then be replaced with what was proposed: e.g. removing the image then adding the new one? What happens when a user tries to do this now?

HarmvZ: I could not find if that replacement functionality was actually implemented. What happens now (in the test at least) is that a new CIV is created with the new interface instead of replacing the old one.

I see now that I actually removed a bit too much here. The first sentence should not have been removed, adding it back now.

PR #157, opened 2024-07-29
Check if current version is the latest version

### LINKED ISSUE #152: Detect when client is out of date
At the moment it is unclear if the user is running an outdated client. It would be good to add a warning on startup if the client is out of date, and an error if an old version is explicitly unsupported. We would probably need a new endpoint on Grand Challenge to do this.

HarmvZ:
Closes #152 

Checks the version and produces a warning on import when a newer version is available on pypi.


jmsmkn (approved):
I was wondering if we should have an API endpoint for this in addition to this? That way we can block versions that are no longer supported and force an upgrade.

HarmvZ:
That's a good idea. So an endpoint on gc that returns the lowest supported version which will make gcapi raise an error if it's version is lower?

jmsmkn:
Indeed!

chrisvanrun:
Eh, something went wrong with drinking my coffee AND my local main wasn't clean =/

chrisvanrun:
Going to close this in favour of an easy cherry-picked re-do.

PR #171, opened 2024-11-05
Add version checking

chrisvanrun:
This is a partial copy of the outdated PR of @HarmvZ who laid the foundations. However, it now also checks the latest supported version.

The original PR had some problems with double running the check and the new check needed access to the `base_url` anyway. So I decided put the checking in the `ClientBase` initialisation but keep it in a separate httpx Client.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -8,7 +8,7 @@ repos:
       - id: mixed-line-ending
       - id: trailing-whitespace
   - repo: https://github.com/asottile/pyupgrade
-    rev: v3.18.0
+    rev: v3.19.0
     hooks:
       - id: pyupgrade
         language: python
@@ -35,7 +35,7 @@ repos:
           - mccabe
           - yesqa
   - repo: https://github.com/pre-commit/mirrors-mypy
-    rev: 'v1.12.0'
+    rev: 'v1.13.0'
     hooks:
       - id: mypy
         additional_dependencies:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -8,7 +8,7 @@ repos:
       - id: mixed-line-ending
       - id: trailing-whitespace
   - repo: https://github.com/asottile/pyupgrade
-    rev: v3.19.0
+    rev: v3.18.0
     hooks:
       - id: pyupgrade
         language: python
@@ -35,7 +35,7 @@ repos:
           - mccabe
           - yesqa
   - repo: https://github.com/pre-commit/mirrors-mypy
-    rev: 'v1.13.0'
+    rev: 'v1.12.0'
     hooks:
       - id: mypy
         additional_dependencies:
--- a/gcapi/check_version.py
+++ b/gcapi/check_version.py
@@ -13,12 +13,14 @@ def check_version(base_url):
     package_name = "gcapi"
 
     current_version = get_version(package_name)
-    with httpx.Client() as client:
-        response = client.get(f"{base_url}/gcapi/")
 
-        json = response.json()
-        latest_version = json["latest_version"]
-        lowest_supported_version = json["lowest_supported_version"]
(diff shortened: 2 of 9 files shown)


gcapi/check_version.py
```diff
@@ -0,0 +1,49 @@
...
+    pass
+
+
+def check_version(base_url):
+    package_name = "gcapi"
+    try:
+        current_version = get_version(package_name)
+        with httpx.Client() as client:
+            response = client.get(f"https://pypi.org/pypi/{package_name}/json")
+            latest_version = response.json()["info"]["version"]
```
jmsmkn, line 18: Now that we have the endpoint we should put this information into our endpoint and save both a call and a dependency on pypi being up.

gcapi/check_version.py
```diff
@@ -0,0 +1,49 @@
...
+from packaging import version
+
+
+class UnsupportedVersionError(Exception):
+    pass
+
+
+def check_version(base_url):
+    package_name = "gcapi"
+    try:
```
jmsmkn, line 14: I don't think the big try except clause is necessary if we do everything on our own servers now.

.pre-commit-config.yaml
```diff
@@ -8,7 +8,7 @@ repos:
       - id: mixed-line-ending
       - id: trailing-whitespace
   - repo: https://github.com/asottile/pyupgrade
-    rev: v3.18.0
+    rev: v3.19.0
```
jmsmkn, line 11: Are these changes relevant to this PR? Would be nicer if these changes were separate.

jmsmkn (commented):
Please fix the tests.

gcapi/check_version.py
chrisvanrun: Good point, how would you handle the versioning on the GC side though? Periodic pull from pypi or manual?

.pre-commit-config.yaml
chrisvanrun: Had a vague error, turns out my pre-commit was very outdated. We don't really need to update these now. Reverted.

gcapi/check_version.py
jmsmkn: It can be manual for now, taken from an environment variable, less dependencies on external systems the better. Don't think it's too much extra work given the frequency of releases. Then we could automate it later in terraform if needed.

chrisvanrun:
(Forgot to push before reassigning.)

Small trick: autouse fixture to mock check_version during tests.

gcapi/check_version.py
```diff
@@ -0,0 +1,41 @@
...
+def check_version(base_url):
+    package_name = "gcapi"
+
+    current_version = get_version(package_name)
+    with httpx.Client() as client:
+        response = client.get(f"{base_url}/gcapi/")
+
+        json = response.json()
+        latest_version = json["latest_version"]
+        lowest_supported_version = json["lowest_supported_version"]
```
jmsmkn, line 21: Removes shadowing of standard library

```suggestion
    current_version = get_version(package_name)

    with httpx.Client() as client:
        response = client.get(f"{base_url}/gcapi/").json()

        api_data = response.json()

        latest_version = api_data["latest_version"]
        lowest_supported_version = api_data["lowest_supported_version"]
```

jmsmkn (approved):

gcapi/check_version.py
```diff
@@ -0,0 +1,43 @@
...
+    pass
+
+
+def check_version(base_url):
+    package_name = "gcapi"
+
+    current_version = get_version(package_name)
+
+    with httpx.Client() as client:
+        response = client.get(f"{base_url}/gcapi/").json()
```
jmsmkn, line 18:
```suggestion
        response = client.get(f"{base_url}/gcapi/")
```

jmsmkn (approved):

PR #167, opened 2024-08-27
Major refactor of helper functions

chrisvanrun:
This PR refactors existing helper functions (i.e. `Client.update_archive_item`) and adds some new ones that make sense.

These are the helper functions now:

- `update_archive_item`
- `add_cases_to_archive` (new)
- `update_display_set` (new)
- `add_cases_to_reader_study`
- `run_external_job`
- `upload_cases` (untouched)

## Proto models

The core of this PR is the 'gcapi/models_proto.py'.

This PR introduces proto-models. These are models of things that are yet to be. Their main function is to separate the creation/saving logic from the validation logic. This allows for quick unit tests to cover corner cases in validation, while simultaneously allowing to validate all values before anything hits the platform.

These 'preflight' validations were, partially, already in place for `add_cases_to_reader_study` but are now ubiquitous for all helper functions.

While refactoring I ran into several issues concerning the adaptation of models, which I fixed as I went. With deadline pressure and random fixes, flaky tests, et cetera this is now a messy and large change set. Sorry!

Below are some descriptions of additional fixes that are included in the PR.

---

### Both `str`, and [`str`] are allowed
Sourcing a CIV from one file no longer requires a one-lengthed list, the helper functions automagically handle both. This also applies to `pathlib.Path`.

---

### `.details` API now accepts an `api_url` keyword argument
Since we are casting to models it made more sense to be able to call the `.details` with api_url than to manually figure out which `gcapi.models` one is to use.

This replaces:

```Python
 gcapi.models.HyperLinkedImage(**client(url=api_url))
 ```
 
 With:
 ```Python
 client.images.details(api_url=api_url)
 ```
 ---

 ### `.create()`/`.update()`/`.partial_update()` -> models

 API calls that support `.create(..)`, `.update(..)`, and `.partial_update(..)` now return `gcapi.models` instead of `dicts`.

---

### Using existing images

Related to the actual pitch:
- https://github.com/DIAGNijmegen/rse-roadmap/issues/305

One can add an existing image as follows (pulled from a test):

```Python
 # First create an image component interface
added_display_sets = client.add_cases_to_reader_study(
    reader_study="reader-study",
    display_sets=[
        {
            "ct-image": [TESTDATA / "image10x10x101.mha"],
        }
    ],
)

display_set_pk = added_display_sets[0]

# ... wait for import to finish

display_set = c.reader_studies.display_sets.detail(pk=display_set_pk)

# Re-use both as a direct image and a CIV
display = display_set.values[0]
[...]

PR #172, opened 2024-11-08
More Data Models

chrisvanrun:
This PR contains some extracted (or more re-achieved) changes from:

- https://github.com/DIAGNijmegen/rse-gcapi/pull/167

When I was doing a major re-haul and unwisely decided that, since I was touching nearly everything, it made most sense to push the changes upon the already large change set.

This PR does two major things:
- Allow object/model retrieval using `api_url` (was done with the direct request, now forced through the APIs)
- Fix resulting deprecation warnings on key access (i.e. `object["pk"]`

It also makes the recurse_call wrapper more robust by outputting the error (`raise from e`) and allowing for retries when `ObjectNotFound` is thrown.

Widespread changeset, but relatively simple changes.

### DIFF
--- a/gcapi/apibase.py
+++ b/gcapi/apibase.py
@@ -146,9 +146,10 @@ class APIBase(Generic[T], Common[T]):
     def detail(
         self, pk=None, api_url=None, **params
     ) -> Generator[T, dict[Any, Any], T]:
-        if all((pk, params)):
-            raise ValueError("Only one of pk or params must be specified")
-
+        if sum(bool(arg) for arg in [pk, api_url, params]) != 1:
+            raise ValueError(
+                "Exactly one of pk, api_url, or params must be specified"
+            )
         if pk is not None or api_url is not None:
             if pk is not None:
                 request_kwargs = dict(path=urljoin(self.base_path, pk + "/"))
--- a/gcapi/apibase.py
+++ b/gcapi/apibase.py
@@ -143,14 +143,20 @@ class APIBase(Generic[T], Common[T]):
             yield from current_list
             offset += req_count
 
-    def detail(self, pk=None, **params) -> Generator[T, dict[Any, Any], T]:
+    def detail(
+        self, pk=None, api_url=None, **params
+    ) -> Generator[T, dict[Any, Any], T]:
         if all((pk, params)):
             raise ValueError("Only one of pk or params must be specified")
 
-        if pk is not None:
+        if pk is not None or api_url is not None:
+            if pk is not None:
+                request_kwargs = dict(path=urljoin(self.base_path, pk + "/"))
+            else:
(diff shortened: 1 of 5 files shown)


gcapi/apibase.py
```diff
@@ -143,14 +143,20 @@ def iterate_all(self, params=None) -> Iterator[T]:
             yield from current_list
             offset += req_count
 
-    def detail(self, pk=None, **params) -> Generator[T, dict[Any, Any], T]:
+    def detail(
+        self, pk=None, api_url=None, **params
+    ) -> Generator[T, dict[Any, Any], T]:
         if all((pk, params)):
             raise ValueError("Only one of pk or params must be specified")
```
jmsmkn, line 150: I think that the combination of `pk` and `api_url` should also be incompatible. Shouldn't we also check that only one of `pk` or `api_url` or `params` are set?

tests/utils.py
```diff
@@ -13,29 +14,41 @@
...
+        last_error = None
         for _ in range(60):
             try:
                 result = func(*args, **kwargs)
                 break
-            except (HTTPStatusError, ValueError):
+            except (
+                HTTPStatusError,
+                ValueError,
+                ObjectNotFound,  # Lagging permissions
```
jmsmkn, line 25:
```suggestion
                ObjectNotFound,  # Lacking permissions
```

tests/utils.py
```diff
@@ -13,29 +14,41 @@
...
+        last_error = None
         for _ in range(60):
             try:
                 result = await func(*args, **kwargs)
                 break
-            except (HTTPStatusError, ValueError):
+            except (
+                HTTPStatusError,
+                ValueError,
+                ObjectNotFound,  # Lagging permissions
```
jmsmkn, line 46:
```suggestion
                ObjectNotFound,  # Lacking permissions
```

jmsmkn (commented):
Looks good, small comments.

tests/utils.py
chrisvanrun: The lagging was intentional: as in, the permissions are not yet set but will soon be. It's to address a race condition that sometimes happens on my (slow) system.

jmsmkn: Ah okay, then I didn't understand the comment at all. Might be better to put that explanation instead.

gcapi/apibase.py
chrisvanrun: Ah yes, I forgot to manually copy this along! But it should really check that only one of them is set.

tests/utils.py
chrisvanrun: Might be my gaming youth shining through here. Always complain about the lag ;-). I've improved the comments.

jmsmkn (approved):

Issue #153, opened 2023-11-28
Uploads to archive items fail when CIV is file type and JSON sent


jmsmkn:
If you have a file type CIV (e.g. `clinical-information-lung-ct`) and try to upload a JSON object directly to that (e.g. `{'age': 60, 'gender': 'FEMALE', 'smoking_status': 'Ex-smoker', 'clinical_category': None, 'regional_nodes_category': None, 'metastasis_category': None}`) it will fail with `ValueError: You can only upload one single file to a clinical-information-lung-ct interface.`.

This is kindof cryptic, as the solution is for the user to write the file to disk first, then send the name or Path object along.

We should detect this and handle it ourselves as the user should need to know about the details of how the CIV is stored within GC in order to upload it. However, we need to be careful as string values could just point to files.

chrisvanrun:
This now works on main.

PR #173, opened 2024-11-15
Add Create Strategies

chrisvanrun:
Part of the large PR:
- https://github.com/DIAGNijmegen/rse-gcapi/pull/167

This PR adds creation strategies for creating interdependent objects that can be used for creating display sets and archive items. Strategies _should_ return an item that can be used to finalize the strategy, the CIV strategies return a `ComponentInterfaceValuePostRequest`.

Only tests for the prep stage of the strategies have been added since testing the actual execution of the strategies is non-trivial and is heavily reliant on the involved integration tests. The strategies are a result of heavy testing and extraction from the helper functions on the original PR.

This PR implements the logic behind the following:

## Allowing both `str`, and `list[str]`

Sourcing a CIV from one file no longer requires a one-lengthed list, the helper functions automagically handle both. This also applies to pathlib.Path.

## RE-use of images / CIVs

Can provide a CIV or Image to the strategies and it will be re-used.


jmsmkn:
I'm aware of this but because it is non-pitch related and still large it has low priority for me, so keeps getting pushed back by higher priority things. It may still be a while.

chrisvanrun:
> I'm aware of this but because it is non-pitch related and still large it has low priority for me, so keeps getting pushed back by higher priority things. It may still be a while.

Check, I thought as much! Given its size and low priority it makes perfect sense. The assignment was a subtle async poke ;-)

gcapi/create_strategies.py
```diff
@@ -0,0 +1,328 @@
...
+        except FileNotFoundError as file_val_error:
+            # Possibly a directly provided value
+            if self.interface.kind.casefold() == "string":
+                # Incorrect paths being uploaded as content to a String
+                # interface can easily bypass detection.
+                # We'll be overprotective and not allow via-value uploads.
+                raise FileNotFoundError(
+                    f"Interface kind {self.interface.kind} requires to be uploaded "
+                    "as a file, replace the value with an existing file path"
+                ) from file_val_error
```
amickan, line 163: I do not understand what's going on here. Why only check for string interfaces?

gcapi/create_strategies.py
```diff
@@ -0,0 +1,328 @@
...
+                )
+            return super().__new__(cls)
+
+        return super().__new__(handler_class)
+
+    def __init__(
+        self,
+        *,
+        source: Union[
+            FileSource, HyperlinkedComponentInterfaceValue, HyperlinkedImage
```
amickan, line 120: Do users have a way to reference a `HyperlinkedComponentInterfaceValue`?

gcapi/create_strategies.py
```diff
@@ -0,0 +1,328 @@
...
+
+    def prepare(self):
+        yield from super().prepare()
+
+        if isinstance(self.algorithm, str):
+            self.algorithm = yield from self.client_api.algorithms.detail(
+                slug=self.algorithm
+            )
+
+        interface_lookup = {ci.slug: ci for ci in self.algorithm.inputs}
```
amickan, line 297: This will need to be updated to work with `interfaces`

gcapi/create_strategies.py
```diff
@@ -0,0 +1,328 @@
...
+
+        for ci in self.algorithm.inputs:
+            if ci.slug not in self.inputs and ci.default_value is None:
```
amickan, line 300: We no longer allow not specifying an interface if it has a default value. All inputs need to be provided.

gcapi/create_strategies.py
chrisvanrun: I see the comments are insufficient, let me explain and see if we can improve them.

The idea is that you can provide both a direct value and path to a file that contains the value. The heuristic is to find out which is which. If it can't find the file based off the value, we assume it is a directly provided value and sent it off to be validated at GC.

However, if someone misstyped the pathname, say, `"/example/valua.jso"` we don't want to push that to something that might readily accept that and we know users almost never check the contents.

This was sketched as a requirement here: https://github.com/DIAGNijmegen/rse-gcapi/issues/153 . 

Ofc we could have an `Anything` kind without any schema that would have the same problem, but the free `String` kinds are a sure recipe for problems.

gcapi/create_strategies.py
chrisvanrun: Yes, via regular gcapi calls that return these. IIRC that is the `DisplaySet`, `ArchiveItem` and `Job` endpoints. Idea is that you can then readily re-use existing CIVs.

amickan: I think those endpoints just return information about the CIVs (a dictionary with image, file etc., along with the CIV pk). I don't think there is a way to get the `api_url` to a `ComponentInterfaceValue` at the moment, we would need to add that on the GC side.

gcapi/create_strategies.py
amickan: Ah ok, my initial feeling was just don't allow them to provide the values directly if it's a file type interface, but having read #153, I understand the issue better. 

I'm still confused as to why you check for the string CI though. This logic is part of the `FileCIVCreateStrategy`, and so `self.interface.kind.casefold()` should never be `string`, right?

amickan: I think what you would want to do when the file can't be found, is check whether the value provided is valid json according to the schema of the interface, but we don't have access to the schema defined for the interface in gcapi, I think?

gcapi/create_strategies.py
chrisvanrun: Urgh, you are right! I'd like to say that shows the age, but the interfaces are pretty new =P

gcapi/create_strategies.py
chrisvanrun: The fun of models.py! =)

https://github.com/DIAGNijmegen/rse-gcapi/blob/0f9b870d0da775221845d3f59ac0449a5678f6d6/gcapi/models.py#L1518

Via DisplaySet model:
https://github.com/DIAGNijmegen/rse-gcapi/blob/0f9b870d0da775221845d3f59ac0449a5678f6d6/gcapi/models.py#L1717

Via:
https://github.com/DIAGNijmegen/rse-gcapi/blob/0f9b870d0da775221845d3f59ac0449a5678f6d6/gcapi/client.py#L149

And finally via:
https://github.com/DIAGNijmegen/rse-gcapi/blob/0f9b870d0da775221845d3f59ac0449a5678f6d6/gcapi/apibase.py#L162

Will end-up being wrapped in that class.

amickan: Yes, I understand that, but my question was more about how the user will have to supply the CIV to the helper functions. Usually for existing objects (i.e. images) we ask the user to provide the api_url to the image. That does not exist for CIVs yet, and would need to be added on the GC side. But mabye you were envisioning this to work differently?

amickan: Actually, on second thought, I don't think we'd want to open up an endpoint for CIVs given they contain sensitive information, and we don't have permissions attached to them directly.

gcapi/create_strategies.py
chrisvanrun: > I'm still confused as to why you check for the string CI though. This logic is part of the FileCIVCreateStrategy, and so self.interface.kind.casefold() should never be string, right?

`String` kind, `File` superkind: this is a typical combo you'd see when using input for say prompting or text reports.

> I think what you would want to do when the file can't be found, is check whether the value provided is valid json according to the schema of the interface, but we don't have access to the schema defined for the interface in gcapi, I think?

Indeed, we do not. Moreover: the `String` kind might simply not have a neat schema and be free form.

gcapi/create_strategies.py
chrisvanrun: Main goal of the helper functions is to pull the specifics of CIVs and their instantiations away from the user.

The rational here, for having the typing here it so to allow the user do (pseudo):

```
ds = client.display_set.detail(pk=899...)
client.display_set.create(values=["image-ci-slug": ds.values[0], "float-ci-slug": 42])
```

Ofc they could also do `ds.values[0].image` which works but this opens the hood sort of already.

Perhaps this typing here isn't fully correct and rather it should be typed at `ImageCIVCreateStrategy`, as that specialisation only accepts that. However, the best strategy here is likely supporting this for both file and values as well: File types should download the input, value types should directly use the value.

Thoughts?

amickan: Ok, it's clear to me now and it makes sense to abstract this away from the user. Ideally this would work for all socket kinds, indeed.

gcapi/create_strategies.py
amickan: Ok, and your reasoning is that strings are the only ones that might not get rejected server-side because there is no schema for most of those, but that for all other kinds there is validation in place and hence it's fine to accept values and not know if it's just a mistyped path? I think I agree with that.

gcapi/create_strategies.py
chrisvanrun: I've added handlers for when providing a `HyperlinkedComponentInterfaceValue` as a source.

So one can now source both "file"/"value" super_kind `SocketValue` directly from existing `DisplaySet`/`ArchiveItem`.

gcapi/create_strategies.py
```diff
@@ -0,0 +1,455 @@
...
+        self.content = value
+
+    def __call__(self):
+        post_request = yield from super().__call__()
+
+        if (
+            isinstance(self.source, HyperlinkedComponentInterfaceValue)
+            and self.source.interface.super_kind.casefold() == "file"
+        ):
+            temp_file = yield from _download_file(
```
chrisvanrun, line 365: I am not really proud of this construct. I think creating a contextmanagers might work and maybe even a buffer-to-buffer where you stream the download directly into the upload.

However, with the hybrid async/sync attributes it becomes a pain to figure out who to create a contextmanager while still being able to support both async and sync.

Core idea here is that it happens by downloading things to disk first, and then streaming from there to prevent a a memory bottleneck (at least for the pure FileSocketValueStrategy). `ValuesSocketValueStrategy` (i.e. when targeting a 'value' kind socket) will generaly not run into memory problems.

Open for suggestions but atm good enough, imho.

chrisvanrun:
Ready for reviewing, however the integration tests have broken, so need a rebase once this is merged:
- https://github.com/DIAGNijmegen/rse-gcapi/pull/188

gcapi/create_strategies.py
chrisvanrun: Note that these parts (i.e. strategy calling) are not tested here. Rather they are part of the integration testing of the next PR. This is done intentionally as to keep this PR within reviewable size. 

I had some quick-and-dirty test scripts running while developing this.

chrisvanrun:
Jeej. CI passes: ready for re-review!

gcapi/create_strategies.py
```diff
@@ -0,0 +1,455 @@
...
+        self.content = cleaned_sources[0]
+
+        # Test load
+        with open(self.content) as fp:
+            json.load(fp)
+
+    def _prepare_from_existing_socket_value(self, socket_value):
+        socket = socket_value.interface
+        if socket.super_kind.casefold() == "file":
+            pass  # Nothing to do here, but we accept it
```
amickan, line 346: Maybe I misunderstood something, but why do we accept file type interfaces to be handled by the `ValueSocketCreateStrategy`?

amickan: Oh no, sorry, pre-first-coffee brain. So this is when someone submits an existing CIV value of a different socket (which could be a file type socket). I would have expected the code to download the file here rather than in `__call__`. 

But in general, isn't it usually a red flag if someone tries to submit socket A value to socket B value? Most of the time, there are schemas attached to our json sockets, and content for socket A is unlikely to be valid for socket B, right?

gcapi/create_strategies.py
amickan: I don't have a better suggestion at the moment, and think this is indeed fine.

gcapi/create_strategies.py
chrisvanrun: Maybe a good guard-rail to have: only allow socket-to-socket transfers. I think that might be wise to add.

gcapi/create_strategies.py
chrisvanrun: These strategies are core for the (new) helper functions. With the helper functions steering clear from users having to understand the way the platform handles CIVs, CIs and different super kinds.

As such, I have no clue where I'd warn the users apart from the documentation. Maybe the docstring on the eventual helper functions?

amickan: Yes, documentation and the docstring of the helper function would do the job, I think...

gcapi/create_strategies.py
chrisvanrun: Done! Now one can only do exact socket-to-socket copying. Also makes the sourcing of that simpler since we can assume things to be of the exact same kind.

amickan (approved):
LGTM!

PR #174, opened 2024-11-15
Add a serialization function to the models.

chrisvanrun:
Tiny addition that will help with doing `create` calls with the `PostRequest` models.

### DIFF
--- a/gcapi/model_base.py
+++ b/gcapi/model_base.py
@@ -1,5 +1,7 @@
 import warnings
 
+from pydantic import RootModel
+
 
 class BaseModel:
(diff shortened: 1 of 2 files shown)


chrisvanrun:
Huh. I could have sworn this didn't work initially. I did a few quick tests and it seems to work just fine.

Strange that the pydantic docs suggest to use the changes in this PR: https://docs.pydantic.dev/latest/concepts/dataclasses/#json-dumping . Maybe that is just a direct use of a singular function instead of `asict` + `json.dumps`. Still, overly complex.

PR #178, opened 2025-01-16
Fix tests on main are broken

chrisvanrun:
Also adds a few minor tweaks to make local testing more efficient.


gcapi/client.py
```diff
@@ -418,7 +418,7 @@ def __init__(
         self._auth_header = _generate_auth_header(token=token)
 
         self.base_url = URL(base_url)
-        if self.base_url.scheme.lower() != "https":
+        if verify and self.base_url.scheme.lower() != "https":
```
chrisvanrun, line 421: If you turn off `verify`, the root certificate is not checked, so using https isn't of much use then either: man in the middle attack is then undetectable.

gcapi/client.py
```diff
@@ -439,7 +439,10 @@ def __getattr__(self, item):
     def validate_url(self, url):
         url = URL(url)
 
-        if not url.scheme == "https" or url.netloc != self.base_url.netloc:
+        if (
```
chrisvanrun, line 442: Instead of a hard check on https, soft check it matches the base_url scheme.

tests/conftest.py
```diff
@@ -22,14 +26,18 @@ def anyio_backend():
 
 @pytest.fixture(scope="session")
 def local_grand_challenge() -> Generator[str, None, None]:
-    local_api_url = "https://gc.localhost/api/v1/"
+    local_api_url = os.environ.get(
+        "LOCAL_API_URL",
+        "https://gc.localhost/api/v1/",
```
chrisvanrun, line 31: I use http://gc.localhost:8000 locally.

jmsmkn (changes requested):
The changes to the tests look good but please can you revert the checks against https checking? The checks are there not protect against MITM - they are to protect against the auth token being sent over HTTP where everyone has access to it unencrypted. I would very much prefer that we maintained these checks as it would otherwise mean that it is very easy to send the users auth token to GC over http which would result in account compromise.

jmsmkn:
BTW not tried it myself but you might be able to run the test server locally with SSL directly https://django-extensions.readthedocs.io/en/latest/runserver_plus.html

chrisvanrun:
> BTW not tried it myself but you might be able to run the test server locally with SSL directly https://django-extensions.readthedocs.io/en/latest/runserver_plus.html#ssl

Good idea. I'll try that route next time I'm working on gcapi locally.

chrisvanrun:
> The changes to the tests look good but please can you revert the checks against https checking? The checks are there not protect against MITM - they are to protect against the auth token being sent over HTTP where everyone has access to it unencrypted. I would very much prefer that we maintained these checks as it would otherwise mean that it is very easy to send the users auth token to GC over http which would result in account compromise.

Reverted. Let's be double sure the users are using https when sending the tokens!

pkcakeout:
Quickly writing it here as well: https://github.com/DIAGNijmegen/rse-cirrus-core/issues/3377 CIRRUS is broken, I suspect due to the ordering thing, maybe? I did not have time to sink my teeth in this yet, needs to wait until Monday.

pyproject.toml
```diff
@@ -24,6 +24,7 @@ pydantic = "*"
 pytest = "*"
 pytest-randomly = "*"
 pytest-cov = "*"
+python-dotenv = "*"
```
jmsmkn, line 27: Do we need a new package for this? I thought that on Grand Challenge we recommend using the poetry plugin for it and then we can read directly from `os.environ.get`?

tests/conftest.py
jmsmkn: Might be good to namespace the env var, `GCAPI_TESTS_LOCAL_API_URL` maybe?

jmsmkn (commented):
Looks good! Couple of minor questions.

pyproject.toml
chrisvanrun: Ah, totally forgot about that combination of poetry + dotenv. I'llfollow that route: way less of an impact, ableit I am surprised by the dependencies it has:

afbeelding

jmsmkn (approved):
Thanks!

PR #187, opened 2025-05-09
Update models + fix some flaky tests

chrisvanrun:


amickan (dismissed):

chrisvanrun:
Huh, intermittent failures with request headers. Let's see if I can reproduce this locally.

chrisvanrun:
Going to re-run the tests for now and introduce a fix via a different PR.

chrisvanrun:
Alas, a re-run still has httpbin.org fail with 503. Going to push the local httpbin fix here after all.

chrisvanrun:
And another flaky test now fails because of the minor shift in race conditions. Fixed: I think.

PR #189, opened 2025-05-16
Update archiveItem helper functions

chrisvanrun:
Hooks the archive_item helper functions into the socket strategies.

The existing `update_archive_item` and the new `add_cases_to_archive` (new: similar level as the `add_cases_to_reader_study`).

Merge note: to keep things reviewable I am limiting the change set per category of helper functions. Still needs the target to be merged though.

Dependent on the GC counter-part:
- https://github.com/comic/grand-challenge.org/pull/4030

### DIFF
--- a/gcapi/client.py
+++ b/gcapi/client.py
@@ -721,7 +721,7 @@ class ClientBase(ApiDefinitions, ClientInterface):
             if interfaces.get(slug):
                 interface = interfaces[slug]
             else:
-                interface = yield from self._fetch_interface(slug)
+                interface = yield from self._fetch_socket_detail(slug)
                 interfaces[slug] = interface
             super_kind = interface.super_kind.casefold()
             if super_kind != "value":
--- a/gcapi/client.py
+++ b/gcapi/client.py
@@ -7,7 +7,7 @@ from io import BytesIO
 from pathlib import Path
 from random import randint
 from time import sleep
-from typing import TYPE_CHECKING, Any, Callable, Optional, Union
+from typing import TYPE_CHECKING, Any, Callable, Optional, Union, cast
 from urllib.parse import urljoin
 
 import httpx
@@ -16,9 +16,11 @@ from httpx import URL, HTTPStatusError, Timeout
 import gcapi.models
 from gcapi.apibase import APIBase, ClientInterface, ModifiableMixin
 from gcapi.check_version import check_version
+from gcapi.create_strategies import Empty, SocketValueCreateStrategy
 from gcapi.exceptions import ObjectNotFound
 from gcapi.retries import BaseRetryStrategy, SelectiveBackoffStrategy
 from gcapi.sync_async_hybrid_support import CapturedCall, mark_generator
+from gcapi.typing import SocketValueSet
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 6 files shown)


gcapi/client.py
```diff
@@ -904,3 +781,218 @@ def add_cases_to_reader_study(
...
+
+            return interface
+        except ObjectNotFound as e:
+            raise ValueError(
+                f"{slug} is not an existing interface. "
+                f"Please provide one from this list: "
+                f"{ref_url}"
+            ) from e
+
+    def _create_socket_value_sets(
```
chrisvanrun, line 920: These methods willalso  play a role in display-set helper functions.

tests/async_integration_tests.py
chrisvanrun, line 1: The tests seem very fragile, so I partially rewrote them.

chrisvanrun:
Time to rebase this.

gcapi/client.py
```diff
@@ -904,3 +781,249 @@ def add_cases_to_reader_study(
 
             res.append(ds.pk)
         return res  # noqa: B901
+
+    def _validate_display_set_values(self, values, interfaces):
```
amickan, line 785: is this meant to be in this PR? It looks it's not getting called, right?

gcapi/client.py
```diff
@@ -904,3 +781,249 @@ def add_cases_to_reader_study(
...
+        creation_kwargs: dict,
+        descriptions: list[SocketValueSetDescription],
+        api: ModifiableMixin,
+    ):
+        interface_cache: dict[str, gcapi.models.ComponentInterface] = {}
+
+        # Firstly, prepare ALL the strategies
+        strategy_collections: list[list[SocketValueCreateStrategy]] = []
+        for description in descriptions:
+            strategy_per_value = []
```
amickan, line 963: Maybe clearer to call this `strategy_per_value_set`. I was initially confused and thought this referred to CIVs.

tests/async_integration_tests.py
```diff
@@ -482,231 +482,114 @@ async def test_add_and_update_file_to_archive_item(local_grand_challenge):
...
                 raise ValueError
             return archive_item
 
-        updated_item = await get_updated_archive_item()
+        item_updated = await get_updated_archive_item()
 
-        updated_civs_by_slug = {
-            item.interface.slug: item for item in updated_item.values
-        }
+        csv_socket_value = item_updated.values[0]
```
amickan, line 522: Are items guaranteed to be sorted this way?

amickan (approved):
LGTM, just 2 small questions.

gcapi/client.py
chrisvanrun: Ah, it moved! I originally started with clearing out some helper functions that are not longer needed to paste all of them. Thought it wiser to re-add all those other than the archive-items so I could scope things a bit to keep this PR smallish.

For clarity I'll move it where it was (above add_cases)

amickan: Ok, so it's needed for some other helper function, but not the ones you're touching here.

tests/async_integration_tests.py
chrisvanrun: I think so ... several other tests rely on. However: better to not assume or rely on this.

gcapi/client.py
chrisvanrun: I've updated a few collection names. Thought I think it IS meant to be a strategy per CIV. Or a strategy that should create CIVs: eventually.

Let me know what you think of the new names.

gcapi/client.py
chrisvanrun: Indeed!

PR #192, opened 2025-05-20
Update DisplaySet helper functions

chrisvanrun:
Part of pitch:
- https://github.com/DIAGNijmegen/rse-roadmap/issues/406

This PR will update the existing helper function `add_cases_to_reader_study` and add a new helper function `update_display_set` (+ tests).

In addition, a deprecation warning will now fire when someone uses `upload_cases`. Imho that function should go since it does upload for answers/archives/archive_item/reader_sets. However, only for images. Which is very confusing. Far better to use the resilient strategies for that via the other helper functions. 

Also: some renaming of "interface" => "socket".

### DIFF
--- a/gcapi/client.py
+++ b/gcapi/client.py
@@ -798,8 +798,9 @@ class ClientBase(ApiDefinitions, ClientInterface):
                 https://grand-challenge.org/reader-studies/i-am-a-reader-study/
 
         display_sets
-            The format for the description of display sets is as follows:
+            The format for the descriptions of display sets are as follows:
 
+            ```python
             [
                 {
                     "slug_0": ["filepath_0", ...],
@@ -811,11 +812,29 @@ class ClientBase(ApiDefinitions, ClientInterface):
                 },
                 ...
             ]
+            ```
 
             Where the file paths are local paths to the files making up a
             single image. For file-kind sockets the file path can only
             reference a single file. For json-kind sockets any value that
-            is valid for the sockets can directly be passed.
+            is valid for the sockets can directly be passed, or a filepath
+            to a file that contain the value can be provided.
+            Existing images on Grand Challenge can be re-used by either
+            passing an API url, or a socket value (display set):
+
+            ```python
+            image = client.images.detail(pk="ad5...")
+            ds = client.reader_study.display_set.detail(pk="f5...")
+            socket_value = ds.values[0]
+
+            display_sets = [
+                {
+                    "slug_0": image.api_url,
+                    "slug_1": socket_value,
+                    "slug_1": socket_value.image.api_url,
+                }
+            ]
+            ```
 
         Returns
         -------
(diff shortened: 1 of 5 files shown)


gcapi/client.py
```diff
@@ -715,103 +728,116 @@ def run_external_job(  # noqa: C901
...
+                    "slug_n": {"json": "value"}
+
+                },
+                ...
+            ]
+
+            Where the file paths are local paths to the files making up a
+            single image. For file-kind sockets the file path can only
+            reference a single file. For json-kind sockets any value that
+            is valid for the sockets can directly be passed.
```
amickan, line 818: Again maybe worth adding an example of how one can pass in an existing CIV?

amickan: Same above for the other helper funcion

gcapi/client.py
```diff
@@ -715,103 +728,116 @@ def run_external_job(  # noqa: C901
...
+
+        reader_study = client.reader_studies.detail(slug="...")
+        items = list(
+            client.reader_studies.display_sets.iterate_all(
+                params={"reader_study": reader_study.pk}
+            )
+        )
+
+        To then add, for example, a PDF report and a lung volume
+        value to the first archive item , provide the interface slugs together
```
amickan, line 750:
```suggestion
        value to the first display set , provide the interface slugs together
```

gcapi/client.py
```diff
@@ -715,103 +728,116 @@ def run_external_job(  # noqa: C901
...
-        study. The format for the list of display sets is as follows:
-        [
-            {
-                "slug_0": ["filepath_0", ...]
-                ...
-                "slug_n": {"json": "value"}
+        This function updates an existing display set with the provided values
+        and returns the updated display set.
 
+        You can use this function, for example, to add metadata to an display set.
```
amickan, line 738:
```suggestion
        You can use this function, for example, to add metadata to a display set.
```

amickan (approved):

amickan:
> In addition, a deprecation warning will now fire when someone uses `upload_cases`. Imho that function should go since it does upload for answers/archives/archive_item/reader_sets. However, only for images. Which is very confusing. Far better to use the resilient strategies for that via the other helper functions.

I agree that it should go eventually. Is uploading images to answers used? We don't have a replacement helper function for that now, but it seems a bit of a weird usecase anyway, right?

chrisvanrun:
> > In addition, a deprecation warning will now fire when someone uses `upload_cases`. Imho that function should go since it does upload for answers/archives/archive_item/reader_sets. However, only for images. Which is very confusing. Far better to use the resilient strategies for that via the other helper functions.
> 
> I agree that it should go eventually. Is uploading images to answers used? We don't have a replacement helper function for that now, but it seems a bit of a weird usecase anyway, right?

The answer part is indeed the only thing that might (?) have merrit. However, I don't think CIRRUS uses that and I have never heard of anyone using it.

gcapi/client.py
chrisvanrun: That is a good suggestion. I've added a quick snippet on images and also made a notion of value-kinds kept in files.

PR #193, opened 2025-05-20
Update JobCreation helper functions

chrisvanrun:
Refactors the `run_external_job` to use the `JobInputsCreateStrategy`.

Also adds a minor recurse call as to account for the (flaky) possibility that a Job is still validating inputs when checking. Any other state is OK-ed.

CI will fail until this is merged on the GC side:
- https://github.com/comic/grand-challenge.org/pull/4040

### DIFF
--- a/gcapi/client.py
+++ b/gcapi/client.py
@@ -706,6 +706,16 @@ class ClientBase(ApiDefinitions, ClientInterface):
                     }
                 ]
 
+            One can also provide a same-socket socket value::
+
+                ai = client.archive_items.detail(pk="f5...")
+                archive_items = [
+                    {
+                        "slug_0": ai.values[0],
+                        "slug_1": ai.values[1],
+                        "slug_2": "some_local_file",
+                    },
+                ]
 
         Returns
         -------
@@ -841,6 +851,17 @@ class ClientBase(ApiDefinitions, ClientInterface):
                     }
                 ]
 
+            One can also provide a same-socket socket value::
+
+                ds = client.reader_study.display_sets.detail(pk="f5...")
+                display_sets = [
+                    {
+                        "slug_0": ds.values[0],
+                        "slug_1": ds.values[1],
+                        "slug_2": "some_local_file",
+                    },
+                ]
+
 
         Returns
         -------
@@ -954,6 +975,7 @@ class ClientBase(ApiDefinitions, ClientInterface):
             reference a single file. For json-kind sockets any value that
             is valid for the sockets can directly be passed, or a filepath
             to a file that contain the value can be provided.
+
             Existing images on Grand Challenge can be re-used by either
             passing an API url, or a socket value (display set)::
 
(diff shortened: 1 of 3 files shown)


gcapi/client.py
```diff
@@ -640,24 +644,34 @@ def upload_cases(  # noqa: C901
...
         the algorithm at
             https://grand-challenge.org/algorithms/corads-ai/
         the slug for this algorithm is "corads-ai".
-        For each input interface defined on the algorithm you need to provide a
-        key-value pair (unless the interface has a default value),
-        the key being the slug of the interface, the value being the
-        value for the interface. You can get the interfaces of an algorithm by calling
+
+        For each input socket defined on the algorithm you need to provide a
+        key-value pair (unless the socket has a default value),
```
amickan, line 663:
```suggestion
        key-value pair,
```

amickan: We no longer allow omitting default values

gcapi/client.py
```diff
@@ -640,24 +644,34 @@ def upload_cases(  # noqa: C901
...
+        calling
             client.algorithms.detail(slug="corads-ai")
-        and inspecting the ["inputs"] of the result.
-        For image type interfaces (super_kind="Image"), you can provide a list of
+
+        and inspecting the ["interfaces"] of the result.
+
+        For image-kind sockets (super_kind="Image"), you can provide a list of
         files, which will be uploaded, or a link to an existing image.
+
```
amickan, line 674: Maybe give an example that reuses an existing CIV and how the user would pass that in?

gcapi/client.py
amickan: I believe the create strategy checks for this correctly, but probably good to double check!

gcapi/client.py
chrisvanrun: Can we actually use existing CIVs? How would that work? Adding a api_url to the CIV?

chrisvanrun:
(quick rebase onto latest merged onto main)

gcapi/client.py
amickan: The new thing that that allows is linking existing non-image files (even though they will still be re-uploaded), but so far the user had to have them locally first, now gcapi does the download for the user. Right?

gcapi/client.py
chrisvanrun: Double checked, and indeed!

gcapi/client.py
chrisvanrun: > I was referring to what you explained in the create strategies PR [here](https://github.com/DIAGNijmegen/rse-gcapi/pull/173#discussion_r2050612685). Did I misunderstand that?

Nope. Indeed. That works. I was thinking how to do the 'run from archive' efficiently via the gcapi. I thought perhaps we could point to existing CIVs directly: but that is not the case atm.

chrisvanrun: I've heavily updated the docstrings, correcting some naming and indentation. The Visual Studio Code within IDE now renders them nicely.

chrisvanrun:
Fingers crossed for the CI.

gcapi/client.py
amickan: What do you mean that is not the case at the moment? A user could loop over all archive items in an archive and then schedule a job for each by supplying the values from the archive item as inputs to the job:
`client.run_external_job(
                algorithm="corads-ai",
                inputs=archive_item["values"]
                )`

amickan (approved):

PR #198, opened 2025-05-22
Migrate from Poetry to uv

chrisvanrun:
Since Grand-Challenge now uses `uv` instead of `poetry` we are also migrating gcapi to do someting similar.

Only caveat here is `POETRY_PYPI_TOKEN_PYPI` that is in the GitHub secrets: can be updated at a later stage.


jmsmkn (changes requested):
I don't think this has been tested as publish does not have a `--build` option. Please test this locally by creating another pypi package rather than having to debug this through CI, review, tag, retry.

chrisvanrun:
Create a PyPi account, and a fresh clone of this repo. Quickly renaming the project to something else and tested locally:

```
uv build
UV_PUBLISH_TOKEN="pypi-A..." uv publish
```

Created a new PyPi project with the correct release, with all the values similar to that found on the `gcapi` PyPi page.

Token, release and project have been removed again and this PR updated.

jmsmkn (approved):
LGTM, for future repos we don't need to change the secret name right?

chrisvanrun:
Going to park this PR, for now: to first quickly release some hotfixes.

Issue #201, opened 2025-05-23
A few release bugs


chrisvanrun:
Quick notes of some errors

chrisvanrun:
## `packaging` deps
Missing package of `packaging`. Context: trying to add gcapi via `uv`, maybe that is a corner case. Better add it to deps anyway.

chrisvanrun:
## Ill-constructed DisplaySetPost model

Production GC is reporting this when runing a quick test script... why is this not showing up locally?

Initial error and hunc is that:

```
@dataclass
class HangingProtocol(BaseModel):
    json_: Any
    title: str
    pk: str
    svg_icon: str
```

Does not actually have a `json_` in it.

First: why does this not show up in local instances?

chrisvanrun:
> I think the only way to cover them in integration tests is to produce all possible objects.

True, not all objects can be practically covered. However, imho the common `Answer` is not done nor a `DisplaySet`.

 I've quickly added those now and it does indeed seem that updating the current model to the following does allow GETting them.

```Python
@dataclass
class Answer(BaseModel):
    answer: Optional[Any]
    api_url: str
    created: str
    creator: str
    display_set: Optional[str]
    pk: str
    question: str
    modified: str
    answer_image: Optional[str]
    last_edit_duration: Optional[str]
    total_edit_duration: Optional[str]
    is_ground_truth: bool
```

A quick play-around with the my local dev also shows that the specs are generated correctly if we expicitly set the `allow_null=True` on those fields. However, like you mentioned: why they do not source this from the underlying model fields in the first place?

I spent some time diving into the DRF source code, but things get complex really quickly. Also found: https://github.com/axnsan12/drf-yasg/issues/310#issuecomment-468879166

Going to throw in the towel and just explicitly set `allow_null`. This seems to have been the solution for a few other fields already.

chrisvanrun:
Latest release should have fixed all above...

PR #202, opened 2025-05-23
Fix `packaging` dependency and `json_` fields

chrisvanrun:
Closses: #201

## Missing package `packaging`
Missed the `packaging` package because it is also a dependency of dev-group dependency `datamodel-code-generator`.

## HangingProtocol models
In practice every reader study has a hanging protocol. This wasn't tested and an unchecked rewrite of 'json' => 'json_' had the model validators have a stroke.


generate_models.py
```diff
@@ -37,6 +37,13 @@ def main() -> int:
             output_model_type=DataModelType.DataclassesDataclass,
             target_python_version=PythonVersion.PY_39,
             input_file_type=InputFileType.OpenAPI,
+            aliases={
+                # Prevent resolving 'json' fields to 'json_'.
+                # This happends because the pydantic BaseModel has a
```
koopmant, line 42:
```suggestion
                # This happens because the pydantic BaseModel has a
```

gcapi/models.py
```diff
@@ -1,12 +1,12 @@
...
 #   filename:  schema.json
-#   timestamp: 2025-05-09T13:40:55+00:00
+#   timestamp: 2025-05-23T11:53:32+00:00
 
 from __future__ import annotations
 
 from decimal import Decimal
 from enum import Enum
-from typing import Any, Optional, Union
+from typing import Any, Dict, List, Optional, Union
```
koopmant, line 9: How did this happen? These don't appear to be used

generate_models.py
```diff
@@ -37,6 +37,13 @@ def main() -> int:
...
+                # deprecated json() function which makes 'json'
+                # technically an invalid field name
+                "json": "json",
```
koopmant, line 45: Seems a bit hacky, because I don't think this is what the aliases were intended for?

Can you point me to where the deprecated json() function is? I couldn't find it.

koopmant (commented):
I'm having a hard time reviewing this, because I don't know how the datamodel_code_generator library works. I'm a bit out of my depth here.

Can you add a test for the fix for the json attribute?

gcapi/models.py
chrisvanrun: They get automatically added, but with the pre-commit formatting they end-up not being used. I've let them in to keep the generated models.py 'pure'.

chrisvanrun:
> Can you add a test for the fix for the json attribute?

Good point. In this case this is indirectly done by adding the creation of a `HangingProtocol` to the test fixtures. Without the fixes the integration tests fail.

jmsmkn (approved):
LGTM

PR #205, opened 2025-05-27
Add Syncronous Wrapper for Async Client

chrisvanrun:
Nice-to-have of pitch:
- https://github.com/DIAGNijmegen/rse-roadmap/issues/406

This adds the wrapper that is not hooked into anything: just adds the wrapper what would allow us to define the client in pure async 'colours'.

What is purposefully left not-implemented is using a wrapped asynchronous client in a asynchronous context: developers should just use the asynchronous client instead.

Maybe a ToDo down the road: try to correct static typing of the wrapped client. For now, at least, MyPy does not complain.

It has some tests that cover (I think) all the ways the `AsyncClient` can be used. 
I opted to use the very basic runtime `asyncio.run` over the static wrapper `asgrif.sync.async_to_sync`. The latter breaks down for two reasons:
- By default adding the `async_to_sync` statically does not handle cross-calling of wrapped methods
- When using `async_to_sync` dynamically it is faced with async generators correctly (e.g. `iterate_all`) it only yields once and then shuts down the generator

### DIFF
--- a/gcapi/sync_support.py
+++ b/gcapi/sync_support.py
@@ -1,0 +1,128 @@
+import asyncio
+import inspect
+import threading
+
+from gcapi.apibase import APIBase
+
+
+def is_async_callable(obj):
+    """Checks for async methods but also hits async __call__ methods"""
+    if inspect.iscoroutinefunction(obj):
+        return True
+    else:
+        try:
+            call_method = obj.__call__
+        except AttributeError:
+            return False
+        else:
+            return inspect.iscoroutinefunction(call_method)
+
+
+def assert_not_in_running_loop():
+    """
+    When in a running asynchronous context, we present an informative error message.
+    """
+    try:
+        event_loop = asyncio.get_running_loop()
+    except RuntimeError:
+        pass
+    else:
+        if event_loop.is_running():
+            raise RuntimeError(
+                "Cannot use the synchronous client while an event loop is running. "
+                "Please use the asynchronous client in an asynchronous context."
+            )
+
+
+class SyncCompatWrapper:
+    """
+    A wrapper for an asynchronous instance that allows synchronous access
+    to its methods.
+
+    Does not work in an already running event loop context. Sorry!
+
+    Usage:
+    ```python
+        async_instance = AsyncClient()
+        sync_instance = SyncCompatWrapper(async_instance=async_instance)
+
+        sync_result = sync_instance.some_async_method()
+    ```
+    """
+
+    def __init__(self, *, async_instance, lock=None):
+        assert_not_in_running_loop()
+
+        self.__async_instance = async_instance
+        self.__wrapping_lock = lock or threading.Lock()
+
+    def __wrap_async_callable(self, /, func):
+        def wrapper(*args, **kwargs):
+            with self.__wrapping_lock:
+                coro = func(*args, **kwargs)
+                return asyncio.run(coro)
+
+        return wrapper
+
+    def __wrap_async_gen_function(self, /, func):
+        def wrapper(*args, **kwargs):
+            aiter = func(*args, **kwargs)
+            while True:
+                try:
+                    sync_func = self.__wrap_async_callable(aiter.__anext__)
+                    yield sync_func()
+                except StopAsyncIteration:
+                    self.__wrap_async_callable(aiter.aclose)()
+                    break
+
+        return wrapper
+
+    def __getattr__(self, name):
+        attr = getattr(self.__async_instance, name)
+
+        if self.__wrapping_lock.locked():
+            # We're in an intentionally asynchronous context,
+            # so we return the attribute
+            return attr
+        elif isinstance(attr, APIBase):
+            return SyncCompatWrapper(
+                async_instance=attr, lock=self.__wrapping_lock
+            )
+        elif is_async_callable(attr):
+            return self.__wrap_async_callable(attr)
+        elif inspect.isasyncgenfunction(attr):
+            return self.__wrap_async_gen_function(attr)
+        elif isinstance(attr, APIBase):
+            return SyncCompatWrapper(
+                async_instance=attr, lock=self.__wrapping_lock
+            )
+        else:
+            return attr
+
+    def __call__(self, *args, **kwargs):
+        if is_async_callable(self.__async_instance):
+            sync_func = self.__wrap_async_callable(self.__async_instance)
+        else:
+            sync_func = self.__async_instance
+
+        return sync_func(*args, **kwargs)
+
+
+class SyncCompatWrapperMeta(type):
+    """
+    A metaclass that wraps the instance of a class in a `SyncCompatWrapper`
+
+    Usage:
+    ```python
+        class MySyncClass(MyAsyncClass, metaclass=SyncCompatWrapperMeta):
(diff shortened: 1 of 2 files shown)


gcapi/sync_support.py
```diff
@@ -0,0 +1,128 @@
...
+from gcapi.apibase import APIBase
+
+
+def is_async_callable(obj):
+    """Checks for async methods but also hits async __call__ methods"""
+    if inspect.iscoroutinefunction(obj):
+        return True
+    else:
+        try:
+            call_method = obj.__call__
```
pkcakeout, line 14:
```suggestion
            _ = obj.__call__
```

gcapi/sync_support.py
```diff
@@ -0,0 +1,128 @@
...
+                async_instance=attr, lock=self.__wrapping_lock
+            )
+        elif is_async_callable(attr):
+            return self.__wrap_async_callable(attr)
+        elif inspect.isasyncgenfunction(attr):
+            return self.__wrap_async_gen_function(attr)
+        elif isinstance(attr, APIBase):
+            return SyncCompatWrapper(
+                async_instance=attr, lock=self.__wrapping_lock
+            )
```
pkcakeout, line 98:
```suggestion
```

Duplicate from above?!

gcapi/sync_support.py
```diff
@@ -0,0 +1,128 @@
...
+            aiter = func(*args, **kwargs)
+            while True:
+                try:
+                    sync_func = self.__wrap_async_callable(aiter.__anext__)
+                    yield sync_func()
+                except StopAsyncIteration:
+                    self.__wrap_async_callable(aiter.aclose)()
+                    break
+
+        return wrapper
```
pkcakeout, line 78: This one is interesting... did you test it? Technically, if you run it as a sync function, you are invoking the `__anext__` and `aclose()` functions in a newly constructed event loop on each iteration, right? I am half-way sure that his is unstable and non-intended. Imagine, for example, that the first async-call causes resources to be created on the event-loop. The generator "expects" resources, like a network socket, to persist throughout the async generator call. However, if you tear down the even loop between each call, you tear down those intermediate resources as well... I suspect that this could outright crash the entire python interpreter (worst case) when you:

- Use a non-builtin event-loop, like `uvloop`, which comes with its own networking implementation
- Invoke a generator that returns chunks from a single httpx-request from grand-challenge api

I do not know if:
- We have such an iterator in the first place? Maybe not... maybe we dodge a bullet in this case.
- ... I will write up a bit more about this in a general comment below.

pkcakeout:
Obviously, the async-generator implementation sent my head spinning. This is partial knowledge about how I understand that python works, likely incomplete and I need to read up more... it is late already though, so I will do that tomorrow. What we really need to test:

Is this stable when calling different kinds of functions repeatedly and also through 3rd-party even loops like `trio` or `uvloop`? I cannot quite model this in my head, I think there could be issues.

## Why there could be issues with 3rd party loops

We use httpx. Httpx requires a working socket subsystem. For async clients it very likely uses the constructors from the `asyncio` libraries to manage the socket connections in a socket pool. The [httpx pool](https://github.com/encode/httpx/blob/6c7af967734bafd011164f2a1653abc87905a62b/httpx/_transports/default.py#L135) is somewhere in here, but I do not 100% get how it works right now. If references to sockets are held as part of the pool and those are constructed through the event loop (both!), this might break if the event-loop itself is not persistent.

I cannot quite imagine that httpx would not control for this though and at least invalidate the pool between requests, so we might be golden still here. But if we have persistent connections across a, for example, async generator, it cannot correct that for us. Then it will fail flat on its face.

3rd party event loops likely reveal this issue, because I suspect that the default Python implementation simply uses the builtin `socket` implementation and references to those objects internally. That means that implicitly, no lifetime dependencies exist between an event loop and the created socket resources when using the builtin event loop (**assumption!**). 3rd party event loops might work differently here, and tie lifetimes or use of created objects to the instance of the event loop itself.

This is definitely a thing, and you can observe faults like these with builtin asyncio when trying to use Task-objects on event loops that are created on different threads, which is not supported. More specifically: Thread 1: create a task, hand it to thread 2. Thread 2: Await that task... this will error out with an error "Task-1 not created on this event loop" - something like that.

## If this is an issue, what to do?

To make async generators work, we likely need to spawn a persistent even loop into a separate thread and use active synchronization with that thread to pull out async-generator values synchronously then and feed them into a synchronous generator... I will not eleborate more. Maybe you, @chrisvanrun have thought about those things or tested them before as well in some capacity?

gcapi/sync_support.py
```diff
@@ -0,0 +1,128 @@
...
+    def __call__(self, *args, **kwargs):
+        if is_async_callable(self.__async_instance):
+            sync_func = self.__wrap_async_callable(self.__async_instance)
+        else:
+            sync_func = self.__async_instance
+
+        return sync_func(*args, **kwargs)
+
+
+class SyncCompatWrapperMeta(type):
```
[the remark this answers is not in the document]
chrisvanrun: The idea is to provide a ready-to-go class that can be used in a simple script. The wrapper needs an per-client instance to store the lock at, as such you can't / shouldn't use `__new__` which is mainly about the class?

Idea is to hide the adapter.

gcapi/sync_support.py
chrisvanrun: I initially spawned the loop myself and destroyed it only when the sync generator yielded its last breath. On a whim I tried using the already existing wrapper: checking if the async gen was closed already. It wasn't so I kept this simplified version in it.

chrisvanrun:
Given the complexity creep this wrapper already has and likely the extra work we'll have with typing. I think maybe we should re-write gcapi to be purely user facing and develop the async components at the CIRRUS side where we don't need to consider any 'user-friendly' constructs. 

Now gcapi is doing two things, which might be the wrong direction. I'd prefer to KISS gcapi so it's easy to extend and adapt it as GC grows / changes.

jmsmkn:
> Given the complexity creep this wrapper already has and likely the extra work we'll have with typing. I think maybe we should re-write gcapi to be purely user facing and develop the async components at the CIRRUS side where we don't need to consider any 'user-friendly' constructs.
> 
> Now gcapi is doing two things, which might be the wrong direction. I'd prefer to KISS gcapi so it's easy to extend and adapt it as GC grows / changes.

Agreed.

pkcakeout:
> and develop the async components at the CIRRUS side

Slight confusion here. Will we then go for a "fully synchronous" gcapi? I thought the move was originally planning to go "fully async"? Just need to know where the wind is blowing...

pkcakeout:
And just as a sanity check: This effectively means that we would drop `gcapi` on the CIRRUS-side in the medium term. Not a big deal technically - we are simply doing different kinds of http requests afterall and CIRRUS can definitely handle that.

But project-structure wise, that is intended? I think so, right. Ultimately, when we ... sometime ... are ever fully operating JavaScript/client-side, there will not be a gcapi either.

chrisvanrun:
> Roger then! Just making sure we are all on the same page! Drop the sync client then... or the async client!

Yes, this is very much a... let's walk in way direction... "oh no, I don't like this route! Let us do a 180 degree turn. In style!".

Issue #164, opened 2024-08-07
Deprecate the asynchronous client


jmsmkn:
Maintaining both is too much work and the codebase is a mess. We should push people towards sync and just support that.

PR #209, opened 2025-06-03
Remove hybrid constructs

chrisvanrun:
Part of:
- https://github.com/DIAGNijmegen/rse-gcapi/issues/164

Removes all the hybrid construct tests and generator-like calls (i.e. `yield from`).

Also includes the minimum required typing fixes to get `mypy` to pass with current not-strict settings.

Will follow-up with a PR that does the full typing.


chrisvanrun:
> LGTM, would be good to avoid the use of `Any`.

Indeed. There are a few constructs that might need to change to facilitate that. Might have been better to keep the return type undefined so it gets picked up later. I'll make sure to remove them in the follow up.

PR #210, opened 2025-06-05
Add algorithm images API

chrisvanrun:
Was missing.  Submitted jobs report back a  hyperlinked `algorithm_image`. This allows one to convert that `api_url` to some more details.


chrisvanrun:
Actually, for filtering on the client side this might not be required since the Job objects already have an api_url to algorithm images.

Still, not a bad addition.

Issue #166, opened 2024-08-15
Typing


jmsmkn:
We tried to type this at some point but ran into trouble. See:

- https://github.com/DIAGNijmegen/rse-gcapi/pull/150/commits/e5b6f47d0548a7795bce4d3d7f6c800eb4015a96
- https://github.com/DIAGNijmegen/rse-roadmap/issues/213

chrisvanrun:
Fixing this on the go as the hybrid constructs are being removed.

It does have me wondering: should we make mypy stricter? Setting `strict=True`, logically results in load of errors but there might be a balance somewhere with these, a subset of strict:

```
no_implicit_optional = True
warn_unused_ignores = True
warn_return_any = True
warn_redundant_casts = True
warn_unused_configs = True
strict_equality = True
```

T.b.h. I think the typing is a great aid but it should not become a hassle in a small code base that is gcapi. I vote we add typing where it makes sense.

jmsmkn:
All detailed in https://mypy.readthedocs.io/en/stable/existing_code.html

TLDR: Turn each of the features on one at a time, fixing the errors as you go, until you [complete the strict set](https://mypy.readthedocs.io/en/stable/existing_code.html#introduce-stricter-options) and can set that.

PR #212, opened 2025-06-07
[Typing] Increase strictness MyPy : refactor strategies

chrisvanrun:
Part of:
 - #166 

This PR adds the following strictness to the mypy config:
```
warn_unused_configs = True
warn_redundant_casts = True
warn_unused_ignores = True
strict_equality = True
check_untyped_defs = True
```

This has led me to delete the `ClientInterface` (just use the `Client` directly since we don't have Async vs Sync anymore).

The create strategies for socket values were sort of spagetti already but with typing it became totally tangled and unreadable. As such I bit the bullet and refactor them in a fall-through pattern: each strategy is tried in turn and the first one without a specific error (`NotSupported`) gets used.

Since the `prepare()` was a construct to facilite the hybrid constructs (`yield from`) I've opted to move those checks to the `__init__` of each strategy: makes way more sense imho.

`UserUploads` are now no longer flagged with their archive_item / display_set parent. Instead the user_upload is set in the POST that updates the archive_item/ display_set: far simpler.

### DIFF
--- a/gcapi/apibase.py
+++ b/gcapi/apibase.py
@@ -159,7 +159,7 @@ class ModifiableMixin(Common):
         elif isinstance(data, dict):
             return {k: self.recurse_model_dump(v) for k, v in data.items()}
         elif is_pydantic_dataclass(type(data)):
-            return RootModel[type(data)](data).model_dump()
+            return RootModel[type(data)](data).model_dump()  # type: ignore
         else:
             return data
 
--- a/gcapi/apibase.py
+++ b/gcapi/apibase.py
@@ -1,10 +1,11 @@
+from __future__ import annotations
+
 import collections
 from collections.abc import Iterator, Sequence
-from typing import Any, Generic, TypeVar, overload
+from typing import TYPE_CHECKING, Generic, TypeVar, overload
 from urllib.parse import urljoin
 
-from httpx import URL, HTTPStatusError
-from httpx._types import URLTypes
+from httpx import HTTPStatusError
 from pydantic import RootModel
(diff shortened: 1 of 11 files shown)


gcapi/apibase.py
```diff
@@ -1,40 +1,20 @@
+from __future__ import annotations
```
chrisvanrun, line 1: This allows to reference Client as `Client` instead of `"Client"`.

amickan:
I think this looks good, although I'm not sure I like the implicit ordering of the strategies that `select_socket_value_strategy` relies on. Might be good to at least add a test to ensure that the current order is not messed with accidentally in the future. 


> `UserUploads` are now no longer flagged with their archive_item / display_set parent. Instead the user_upload is set in the POST that updates the archive_item/ display_set: far simpler.

I'm a bit fuzzy on the implementation of the helper functions but does this still need an update to those helper functions to work properly? I guess not since the all tests pass.

chrisvanrun:
Thanks for having a look!

> I'm a bit fuzzy on the implementation of the helper functions but does this still need an update to those helper functions to work properly? I guess not since the all tests pass.

No. The only thing that really changed from the perspective of the helper functions is that they no longer need to ignore the `Empty` creation returns: they no longer occur.

As for ensuring a test: do you recon a simple content test on `_strategy_registry` would be OK? That way we have a canary if something goes wrong and know it is intentional.

chrisvanrun:
> As for ensuring a test: do you recon a simple content test on _strategy_registry would be OK? That way we have a canary if something goes wrong and know it is intentional.

Added an explicit test for the ordering

Issue #218, opened 2025-08-08
CI Flakey


jmsmkn:
Get an almost daily error on one of the builds.

chrisvanrun:
Three calls to the end-point. Race condition as `average_duration` being set during these calls. 

 

> >       assert by_pk == by_slug == by_api_url
>   E       AssertionError: assert Algorithm(api...able=None)])]) == Algorithm(api...able=None)])])
>   E         
>   E         Omitting 8 identical items, use -vv to show
>   E         Differing attributes:
>   E         ['average_duration']
>   E         
>   E         Drill down into differing attribute average_duration:
>   E           average_duration: 0.052943 != None
>   E         Use -v to get more diff

PR #220, opened 2025-08-26
Fix race condition by soley comparing the primary key.

chrisvanrun:
Related: #218

No fully sure if this will fix all the flakeyness but it is a good start.


jmsmkn (approved):
LGTM, thanks!

PR #229, opened 2025-10-13
Add Mkdocs documentation

chrisvanrun:
This PR adds some much requested documentation to gcapi. It uses Google-style docstrings. This is not only the default, but going Numpy made some crossreferencing fail.

## Code documentation
It splits the code documentation in three parts:
- Client (e.g. `__init__`, `__call__`, and convienience functions)
- API (e.g. `.images`, `.archive_items`, etc)
- Models (e.g. `Archive`)

Most of it is nicely crosslinked now.

## Docs from GC
The docs from GC are adapated and placed under 'Usage', a few things under 'Getting Started'

## Live preview
One can preview the docs rendering by running:
`uv run mkdocs serve`

<details><summary>Some screenshots</summary>

### Light mode
afbeelding

### Dark mode
afbeelding

afbeelding

afbeelding

</details>

## Minor code changes
The docstrings rendering had some warnings about missing typing here and there. Instead of adding the typing to the doctrings I prefer them to be pulled from the actual code typing. Adding the code typing led to some complaints by the current MyPy strictness: which had me change a few minor things in the code. Most noticeably, adding a TypeVar to `ModifiableMixin` and merging the `Common` with the `APIBase`.

This actually fixed some bugs too with the incorrect models set at one or two places.

### DIFF
--- a/gcapi/client.py
+++ b/gcapi/client.py
@@ -91,9 +91,7 @@ class ImagesAPI(APIBase[gcapi.models.HyperlinkedImage]):
             elif url is not None:
                 image = self.detail(api_url=url)
             else:
-                image = self.detail(
-                    **params
-                )  # Ensure params is passed correctly
+                image = self.detail(**params)
 
             files = image.files
 
@@ -206,11 +204,11 @@ class ReaderStudiesAPI(APIBase[gcapi.models.ReaderStudy]):
         Get ground truth data for a specific case in a reader study.
 
         Args:
-            pk (str): Primary key of the reader study.
-            case_pk (str): Primary key of the case.
+            pk: Primary key of the reader study.
+            case_pk: Primary key of the case.
 
         Returns:
-            dict: Ground truth data for the specified case.
+            Ground truth data for the specified case.
         """
         return self._client(
             method="GET",
@@ -238,10 +236,10 @@ class AlgorithmJobsAPI(
         Get algorithm jobs filtered by input image.
 
         Args:
-            pk (str): Primary key of the input image to filter jobs by.
+            pk: Primary key of the input image to filter jobs by.
 
         Yields:
-            HyperlinkedJob: Algorithm job instances that use the specified input image.
+            Algorithm job instances that use the specified input image.
         """
         yield from self.iterate_all(params={"image": pk})
 


gcapi/apibase.py
```diff
@@ -114,15 +158,36 @@ def iterate_all(self, params=None) -> Iterator[T]:
...
+        """
         if sum(bool(arg) for arg in [pk, api_url, params]) != 1:
             raise ValueError(
                 "Exactly one of pk, api_url, or params must be specified"
             )
         if pk is not None or api_url is not None:
             if pk is not None:
                 request_kwargs = dict(path=urljoin(self.base_path, pk + "/"))
-            else:
+            elif api_url is not None:
```
HarmvZ: No functional change

gcapi/client.py
```diff
@@ -898,22 +1091,51 @@ def add_cases_to_archive(
...
         """
         This function takes an archive slug or model and a list of archive item
         descriptions and creates the archive item to be used on the platform.
 
-        Parameters
-        ----------
-        archive
-            slug for the archive (e.g. `"i-am-an-archive"`). You can find this
-            readily in the URL you use to visit the archive page:
+        ??? tip "Re-using existing images"
```
HarmvZ: Is there not a way to deduplicate these comments so that this information is only in a single place, but correctly reused for the different functions in the docs?

HarmvZ (commented):
Nice job on this! BTW I did not read all the doc text, I mostly just looked at the functional changes.

gcapi/client.py
```diff
@@ -66,7 +91,9 @@ def download(
             elif url is not None:
                 image = self.detail(api_url=url)
             else:
-                image = self.detail(**params)
+                image = self.detail(
+                    **params
+                )  # Ensure params is passed correctly
```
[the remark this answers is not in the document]
chrisvanrun: I think it was indeed a late-evening todo I added. Removed!

gcapi/client.py
HarmvZ: Right, it is just unfortunate if anything in this changes and someone forgets to update it in all places. But okay your logic makes sense.

HarmvZ (approved):

PR #230, opened 2025-10-17
Dicom create strategy

koopmant:


chrisvanrun:
> Closed because we will replace strategies with explicit inputs

Maybe, not sure how that will eventualy look.

I think it will still be create strategies but the way they get chosen might be more explicit. It is currently a fall-through that has some false-postives because of the non-explicit input.

Might also be a long if-else switch instead of a fall through, with explicit inputs that might be a large but OK construct. With implicit inputs it was waaaaay to complex.

PR #232, opened 2025-10-20
Remove deprecated upload_cases

chrisvanrun:
Had the deprecation warning for a while now. Time to go.


chrisvanrun:
Ehm, might have been a good idea to ALSO remove the tests =/

PR #237, opened 2025-10-28
Trim deps

jmsmkn:


chrisvanrun (approved):
Thanks! I wasn't looking forward to test these one-by-one.

PR #239, opened 2025-10-31
Add SocketValueSpec to strategies

chrisvanrun:
This PR adds a `SocketValueSpec` which is to replace the list of dictionaries that currently describes the proto socket values.

The PR also applies them to the strategies and rewrites the strategies to handle those. I've opted to keep the strategy pattern as this is the most flexible. There is an option now to go for a somewhat deep if-else, but it would pull away the context for each of the strategies from the strategies themselves.

I've opted to use a feature branch approach to this, as this requires a host of small changes and asking anyone to review the very wide changes in one go is unreasonable.

This does entail that until the next PR, most tests and pre-commits will be broken (those relevant are ensured to pass locally though!).

### DIFF
--- a/gcapi/create_strategies.py
+++ b/gcapi/create_strategies.py
@@ -97,8 +97,12 @@ class SocketValueSpec:
 
 
 def clean_file_source(
-    source: Any, *, maximum_number: int | None = None
+    source: list[Path | str] | UnsetType, *, maximum_number: int | None = None
 ) -> list[Path]:
+
+    if isinstance(source, UnsetType):
+        raise ValueError("No files provided")
+
     # Ensure we are handling a list
     sources = [source] if not isinstance(source, list) else source
 
--- a/gcapi/typing.py
+++ b/gcapi/typing.py
@@ -1,14 +1,6 @@
-from pathlib import Path
 from typing import Any, Union
 
 import gcapi.models
-
-FileSource = Union[
-    Path,
-    list[Path],
-    str,
-    list[str],
-]
 
 SocketValueSetDescription = dict[str, Any]
 


gcapi/create_strategies.py
```diff
@@ -10,27 +11,93 @@
...
+            raise ValueError(
+                f"Only one source can be specified, but got: {', '.join(sources)}. "
+                f"Please specify only one of the available source fields."
+            )
+
+        if len(sources) == 0:
+            raise ValueError(
+                "At least one source must be specified. "
+                "Please provide one of the available source fields."
+            )
```
koopmant: I've just done something similar on the GC side. Do we have to validate this on both sides? Can we not rely on the GC side only?

gcapi/create_strategies.py
```diff
@@ -10,27 +11,93 @@
...
+                continue
+            if field_value is not Unset:
+                return field_name
+        raise RuntimeError("No source field is set, this should not happen.")
+
+    def __repr__(self) -> str:
+        set_field_name = self.get_set_field_name()
+        set_field_value = getattr(self, set_field_name)
+        return (
+            f"{self.__class__.__name__}({self.socket_slug=}, "
```
koopmant: This results in `SocketValueSpec(self.socket_slug='foo', ...)` which is not valid. 
```suggestion
            f"{self.__class__.__name__}(socket_slug={self.socket_slug}, "
```

gcapi/create_strategies.py
```diff
@@ -10,27 +11,93 @@
...
+            f"{set_field_name}={set_field_value!r})"
+        )
 
 
 def clean_file_source(
-    source: FileSource, *, maximum_number: int | None = None
+    source: Any, *, maximum_number: int | None = None
```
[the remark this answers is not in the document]
chrisvanrun: Ah, good catch, this was giving me errors on the IDE but not yet caught because pre-commit is having a field day with the not-yet fixed helper functions.

I've updated it to reflect the corrected typing.

gcapi/create_strategies.py
koopmant: Hmm, there are some checks on the GC side to avoid that as well. But I'd have to check if we use the intended pathways, so to speak. Either way, there is no harm in being carefull on both sides I suppose.

chrisvanrun:
> Ah OK, different order. Got it.

Yes, I thought since the end-points weren't done yet I might as well give this rewrite a shot =)

PR #240, opened 2025-10-31
Refactor helper functions to work with SocketValueSpec

chrisvanrun:
Second PR of the feature that adds the `SocketValueSpec`.

This PR deprecates the functions:
- `add_cases_to_archive`
- `add_cases_to_reader_study`
- `run_external_job`

And adds these replacements, respectively:
- `add_case_to_archive`
- `add_case_to_reader_study`
- `start_algorithm_job`

Note the idiosyncrasies in when to supply the `api_url`, when the `pk` and when the `slug`. The functions are all slug based and I've added a few caches to efficiently handle these transformations.

This originally wasn't a problem since one typically called the add_cases handled the loop and we did the lookup only once, just before we looped.

IMHO the information we use the caches for is static enough to last the whole Client life-cycle so they can live on the Client instantiation level.
 
 Tests are known not to pass.
 
 ### Follow up PRs
One follow-up PR will apply these to the lose documentation as well.

The one after (and likely last) will fix the integration tests.

### DIFF
--- a/gcapi/client.py
+++ b/gcapi/client.py
@@ -705,7 +705,7 @@ class Client(httpx.Client, ApiDefinitions):
 
         ??? tip "Re-using existing images"
             Existing images on Grand Challenge can be re-used by either
-            passing an API url, or a socket value (from a display set):
+            passing an API url, or an existing socket value (from a display set):
 
             ```Python
             from gcapi import SocketValueSpec
@@ -716,6 +716,7 @@ class Client(httpx.Client, ApiDefinitions):
 
             inputs = [
                 SocketValueSpec(socket_slug="slug-0", existing_image_api_url=image.api_url),
+                SocketValueSpec(socket_slug="slug-0", existing_socket_value=socket_value),
                 SocketValueSpec(socket_slug="slug-2", existing_image_api_url=socket_value.image),
             ]
             ```
@@ -747,7 +748,7 @@ class Client(httpx.Client, ApiDefinitions):
 
         Returns:
             The newly created Job (post) object. Note that not all inputs will
-                be immediately available therein until the background processing has
+                be immediately available until the background processing has
                 completed.
         """
 
@@ -801,7 +802,7 @@ class Client(httpx.Client, ApiDefinitions):
             client.update_display_set(
                 display_set_pk=items[0].pk,
                 values=[
-                    SocketValueSpec(socket_slug="report", files=["report.pdf"]),
+                    SocketValueSpec(socket_slug="report", file="report.pdf"),
                     SocketValueSpec(socket_slug="lung-volume", value=1.9),
                 ],
             )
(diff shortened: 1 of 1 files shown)


chrisvanrun:
Parking this changeset until after the alt PR gets accepted.

gcapi/client.py
```diff
@@ -948,206 +925,191 @@ def update_archive_item(
...
+
             client.update_archive_item(
-                archive_item_pk=items[0].id,
-                values={
-                    "report": [...],
-                    "lung-volume": 1.9,
-                },
+                archive_item_pk=items[0].pk,
+                values=[
+                    SocketValueSpec(socket_slug="report", files=["report.pdf"]),
```
koopmant:
```suggestion
                    SocketValueSpec(socket_slug="report", file="report.pdf"),
```

gcapi/client.py
```diff
@@ -688,105 +698,89 @@ def run_external_job(
...
 
-            archive_items = [
-                {
-                    "slug_0": image.api_url,
-                    "slug_1": socket_value,
-                    "slug_2": socket_value.image.api_url,
-                }
+            inputs = [
+                SocketValueSpec(socket_slug="slug-0", existing_image_api_url=image.api_url),
+                SocketValueSpec(socket_slug="slug-2", existing_image_api_url=socket_value.image),
```
koopmant: is this valid, passing socket_value.image as image_api_url?

gcapi/client.py
```diff
@@ -798,134 +792,117 @@ def update_display_set(
...
         """
-        This function takes an reader-study slug or model and a list of display-set
-        descriptions. It then creates the display-sets for the reader study.
+        This function takes a reader-study slug and a list of socket value specs.
+        It then creates a single display set for the reader study.
 
         ??? tip "Re-using existing images"
             Existing images on Grand Challenge can be re-used by either
-            passing an API url, or a socket value (display set):
+            passing an API url, or a socket value (from a display set):
```
koopmant:
```suggestion
            passing an API url, or an exisiting socket value:
```

gcapi/client.py
```diff
@@ -798,134 +792,117 @@ def update_display_set(
...
+            from gcapi import SocketValueSpec
+
+
+            image = client.images.detail(pk="ad5...")
+            ds = client.reader_studies.display_sets.detail(pk="f5...")
+            socket_value = ds.values[0]
+
+            values = [
+                SocketValueSpec(socket_slug="slug-0", existing_image_api_url=image.api_url),
+                SocketValueSpec(socket_slug="slug-2", existing_image_api_url=socket_value.image),
```
koopmant: same suggestion as above

gcapi/client.py
koopmant: good to know!

PR #241, opened 2025-10-31
Add SocketValueSpec to strategies - Alternative

chrisvanrun:
This PR adds a `SocketValueSpec` which is to replace the list of dictionaries that currently describes the proto socket values.

The PR also applies them to the strategies and rewrites the strategies to handle those.

After discussing the complexity with @koopmant we've opted to let the registry pattern go and instead accept a large if-else. This entails that the strategies have been reduced to their core: a pre-flight stage (via `__init__`) and an active let-us-hit-gc phase (via `__call__`).

I've opted to use a feature branch approach to this, as this requires a host of small changes and asking anyone to review the very wide changes in one go is unreasonable.

This does entail that until the third PR from this one, most tests and pre-commits will be broken (those relevant are ensured to pass locally though!).

### DIFF
--- a/gcapi/create_strategies.py
+++ b/gcapi/create_strategies.py
@@ -598,15 +598,17 @@ class JobInputsCreateStrategy(BaseCreateStrategy):
         # Find a matching interface
         matching_interface = None
 
-        input_keys = {spec.socket_slug for spec in inputs}
+        input_socket_slugs = {spec.socket_slug for spec in inputs}
 
         best_matching_sockets_count = 0
         best_matching_interface = None
         for interface in self.algorithm.interfaces:
-            interface_keys = {socket.slug for socket in interface.inputs}
-            matching_count = len(input_keys & interface_keys)
-            if matching_count == len(interface_keys):
-                # All input keys are present in the interface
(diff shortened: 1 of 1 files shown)


gcapi/create_strategies.py
```diff
@@ -10,49 +11,114 @@
...
+        value: Direct source value to set (can be None)
+        files: List of source file paths for file/image uploads
+        existing_image_api_url: An API URL of an existing image to reuse
+        existing_socket_value: Existing socket value that can be reused
+    """
+
+    socket_slug: str
+
+    value: Any = Unset
+    files: list[str | Path] | UnsetType = Unset
```
koopmant: I was wondering if we could also add `file` for the case of `maximum_number == 1`, since it feels a bit silly to put one file in a list. But I'm not sure if it's worth it. I'll leave that up to you to decide.

chrisvanrun: Implemented this by some very straight forward wrapping in the `SocketValueSpec` itself.

koopmant (approved):
LGTM!

PR #242, opened 2025-11-04
SocketValueSpec: update and fix integration tests

chrisvanrun:
Second to final PR that adds the SocketValueSpec.

Following some discussion about integration tests I have refactored them. In short: the integration tests now either only read or write without checking for consequences for the latter. This removes a large part of the testing utilities since we no longer probe the integration GC instance if things have imported. As long as the response from GC is OK we consider the create/update test to pass.

In addition, the integration tests have been expanded to also check the reusing of images and socket values as a whole for each of the main super kinds.

All tests and linters should now pass.

Final PR will be a rewrite of the docs examples to use the SocketValueSpecs.

### DIFF
--- a/gcapi/client.py
+++ b/gcapi/client.py
@@ -22,7 +22,7 @@ from gcapi.create_strategies import (
     SocketValueSpec,
     select_socket_value_strategy,
 )
-from gcapi.exceptions import SocketNotFound
+from gcapi.exceptions import ObjectNotFound, SocketNotFound
 from gcapi.retries import BaseRetryStrategy, SelectiveBackoffStrategy
 from gcapi.transports import RetryTransport
 from gcapi.typing import SocketValuePostSet


chrisvanrun:
The tests ran in this order:
```
tests/integration_tests.py::test_reuse_existing_socket_values PASSED
....
tests/integration_tests.py::test_reuse_existing_images PASSED
```

The existing socket value targets the same image in the display set. The tests run synchronously and on the GC side it seems like permissions are handled synchronously as well. A head scratcher, to be sure.

tests/integration_tests.py
```diff
@@ -332,428 +281,229 @@ def test_auth_headers_not_sent(local_httpbin):
...
+    display_set = c.reader_studies.display_sets.detail(
+        pk="14909328-7a62-4745-8d2a-81a5d936f34b"
     )
 
-    assert len(added_display_sets) == len(display_sets)
+    assert len(display_set.values) == 4, "Sanity check"
+    assert display_set.values[0].interface.super_kind.casefold() == "file"
+    assert display_set.values[1].interface.super_kind.casefold() == "value"
+    assert display_set.values[2].interface.super_kind.casefold() == "file"
+    assert display_set.values[3].interface.super_kind.casefold() == "image"
```
koopmant: Maybe check the slugs here instead. That would make it easier to read. Although I don't really know why you are checking this at all.

koopmant (commented):
LGTM! 

One comment. 

One question:
The tests no longer check that the values have actually updated. That is because the API shouldn't need to test this?

PR #252, opened 2025-11-14
Dedicated prep function

chrisvanrun:
This moves the actual prep from the __init__ to a `prepare` function. This allows for some more testing.

Generally, when used as a context manager: the prepare is automatically employed via `__enter__`.


koopmant:
> I am not sure if I follow. Isn't that what is already done in the DICOMImageSetFileCreateStrategy?

Hmm, I'm not sure what I was thinking. Maybe I was confused by the other prep implementations.

